PR #1, opened 2020-06-19
Base java

TheGuyWhoCodes:
## Majors
 - Adds in pom file and basic servlet file per the walkthrough

## Minors
 - .idea folder for Intellij users


.idea/compiler.xml
```diff
@@ -0,0 +1,16 @@
...
+    <annotationProcessing>
+      <profile name="Maven default annotation processors profile" enabled="true">
+        <sourceOutputDir name="target/generated-sources/annotations" />
+        <sourceTestOutputDir name="target/generated-test-sources/test-annotations" />
+        <outputRelativeToContentRoot value="true" />
+        <module name="portfolio" />
+      </profile>
+    </annotationProcessing>
+    <bytecodeTargetLevel>
+      <module name="portfolio" target="1.8" />
```
henryksloan, line 13: "portfolio" appears here twice. I'm not totally sure what these values are used for, but I assume we should replace them with "mediaphile".

.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectModuleManager">
+    <modules>
+      <module fileurl="file://$PROJECT_DIR$/portfolio.iml" filepath="$PROJECT_DIR$/portfolio.iml" />
```
henryksloan, line 5: Same as above: The file "portfolio.iml" and its reference here should be changed.

pom.xml
```diff
@@ -0,0 +1,43 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.google.sps</groupId>
+  <artifactId>portfolio</artifactId>
```
henryksloan, line 7: mediaphile

.idea/compiler.xml
TheGuyWhoCodes: yessir, forgot to change it in the original pom file, intellij generates based off that.

TheGuyWhoCodes:
good comments henry, just fixed em up!

PR #3, opened 2020-06-19
Search Query for Movies + Key Config

TheGuyWhoCodes:
## Majors
 - Implements Search query (`/movies/search`) endpoint to search for movies
 - Adds KeyConfig class that reads config file to get api endpoint, secures it from being on Github

### DIFF
--- a/src/main/java/com/google/sps/KeyConfig.java
+++ b/src/main/java/com/google/sps/KeyConfig.java
@@ -1,0 +1,27 @@
+package com.google.sps;
+
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * @author chris
+ * date: 6/19/2020
+ */
+public class KeyConfig {
+    private static final Properties properties = new Properties();
+
+    static {
+        try {
+            InputStream input = KeyConfig.class.getClassLoader().getResourceAsStream("config.properties");
+            properties.load(input);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    public static String MOVIE_KEY = properties.getProperty("movie_apikey");
+
+}
--- a/src/main/java/com/google/sps/servlets/movie/SearchMovieDatabase.java
+++ b/src/main/java/com/google/sps/servlets/movie/SearchMovieDatabase.java
@@ -1,0 +1,50 @@
+package com.google.sps.servlets.movie;
+
+import com.google.gson.Gson;
+import com.google.sps.KeyConfig;
+import info.movito.themoviedbapi.TmdbApi;
+import info.movito.themoviedbapi.TmdbSearch;
+import info.movito.themoviedbapi.model.core.MovieResultsPage;
+import org.json.simple.JSONObject;
+
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/sps/servlets/movie/SearchMovieDatabase.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * @author chris
+ * date: 6/19/2020
+ */
+@WebServlet("/movies/search")
+public class SearchMovieDatabase extends HttpServlet {
```
henryksloan, line 21: I think MovieSearchServlet would be a better name. The word order parallels /movies/search, and having Servlet in the name will be good for consistency and clarity.

src/main/java/com/google/sps/servlets/movie/SearchMovieDatabase.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * doGet() handles search queries to tmdb database.
+     * @param request: a request may have the following query params: query, pageNumber
+     * @param response: a json object returning pagination info and results
+     * @throws IOException:
+     */
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json;");
```
henryksloan, line 35: Should we make JSON responses like these UTF-8? Not sure, but I suspect we'll run into special characters and other writing systems.

src/main/java/com/google/sps/servlets/movie/SearchMovieDatabase.java
```diff
@@ -0,0 +1,50 @@
...
+import org.json.simple.JSONObject;
+
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * @author chris
```
henryksloan, line 17: Should this name be capitalized? I'm not familiar with the standards for docs generation.

src/main/java/com/google/sps/KeyConfig.java
```diff
@@ -0,0 +1,27 @@
+package com.google.sps;
+
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * @author chris
```
henryksloan, line 10: Capitalize if you think it would be appropriate, otherwise ignore

src/main/java/com/google/sps/servlets/movie/SearchMovieDatabase.java
TheGuyWhoCodes: nah, just takes in your computer username for the $author variable in intellij, is fine

src/main/java/com/google/sps/servlets/movie/SearchMovieDatabase.java
TheGuyWhoCodes: i agree, switching content type

src/main/java/com/google/sps/servlets/movie/SearchMovieDatabase.java
TheGuyWhoCodes: a lot better of name, thanks for this!

TheGuyWhoCodes:
oh lol that null == variable thing is semantically the same to the compiler as variable == null, just stylistic from my last internship

PR #4, opened 2020-06-22
Movie details

TheGuyWhoCodes:
## Majors
 - Add "Movie Details" servlet, users will pass in a movie id and get metadata related to the movie

## Minors
 - Adds in template for config.properties
 - Update gitignore to match for any .properties files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,5 @@ npm-debug.log
 *.stackdump
 scripts/docs.angularjs.org-firebase/functions/content
 /firebase.json
+
+config.properties
--- a/.gitignore
+++ b/.gitignore
@@ -90,4 +90,4 @@ npm-debug.log
 scripts/docs.angularjs.org-firebase/functions/content
 /firebase.json
 
-config.properties
+*.properties
--- a/.gitignore
+++ b/.gitignore
@@ -89,5 +89,3 @@ npm-debug.log
 *.stackdump
 scripts/docs.angularjs.org-firebase/functions/content
 /firebase.json
-
-*.properties
--- a/pom.xml
+++ b/pom.xml
@@ -68,6 +68,13 @@
       <artifactId>themoviedbapi</artifactId>
       <version>1.10</version>
     </dependency>
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
+      <version>1.10.19</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -62,13 +62,16 @@
       <artifactId>appengine-api-1.0-sdk</artifactId>
       <version>1.9.59</version>
     </dependency>
-
+<!--    This is a wrapper for The Movie DB API. Here is the github: https://github.com/holgerbrandl/themoviedbapi-->
+<!--    Here is the API Docs: https://developers.themoviedb.org/-->
     <dependency>
       <groupId>info.movito</groupId>
       <artifactId>themoviedbapi</artifactId>
       <version>1.10</version>
     </dependency>
 
+<!--    Mockito is a testing library, allows for easy mocking and stubbing of Java methods-->
+<!--    here is a good article talking about it: https://www.vogella.com/tutorials/Mockito/article.html-->
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-all</artifactId>
--- a/src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
+++ b/src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
@@ -34,33 +34,27 @@ public class MovieDetailsServlet extends HttpServlet {
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
         response.setContentType("application/json; charset=utf-8");
 
-        Integer id = null;
+        Integer movieId = parseInt(request.getParameter("id"));
 
-        if(null != request.getParameter("id")) {
-            id = parseInt(request.getParameter("id"));
-            if(null == id) {
-                response.sendError(HttpServletResponse.SC_BAD_REQUEST);
-                return;
-            }
-        } else {
+        if (movieId == null) {
             response.sendError(HttpServletResponse.SC_BAD_REQUEST);
             return;
         }
 
         // Uses null to default language to en-US
-        MovieDb queryResponse = moviesQuery.getMovie(id, null);
+        MovieDb queryResponse = moviesQuery.getMovie(movieId, null);
 
         response.getWriter().println(gson.toJsonTree(queryResponse));
     }
 
     /**
      * parseInt method is used to parse an integer from a string (from a url param)
-     * @param urlParam: the url param to convert from string -> integer
+     * @param s: the url param to convert from string -> integer
      * @return either an integer if the string is a valid integer, or null if it can't be parsed
      */
-    private Integer parseInt(String urlParam) {
+    private Integer parseInt(String s) {
         try {
-            return Integer.parseInt(urlParam);
+            return Integer.parseInt(s);
         } catch(Exception e) {
             return null;
         }
--- a/src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
+++ b/src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
@@ -25,6 +25,14 @@ public class MovieDetailsServlet extends HttpServlet {
     private TmdbMovies moviesQuery = new TmdbMovies(new TmdbApi(KeyConfig.MOVIE_KEY));
     private Gson gson = new Gson();
 
+    /**
+     * the doGet() for this servlet will return the user movie details provided a movie id and get
+     * movie meta data, if no id is present in the request, the servlet will return a 400. if the query
+     * can't find the movie, it'll return a 500.
+     * @param request: needs a movie id query parameter
+     * @param response: returns a MovieDb object, including movie metadata.
+     * @throws IOException
+     */
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
         response.setContentType("application/json; charset=utf-8");
@@ -39,6 +47,7 @@ public class MovieDetailsServlet extends HttpServlet {
             response.sendError(HttpServletResponse.SC_BAD_REQUEST);
             return;
         }
+
         // Uses null to default language to en-US
         MovieDb queryResponse = moviesQuery.getMovie(id, null);
 
--- a/src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
+++ b/src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
@@ -1,0 +1,47 @@
+package com.google.sps.servlets.movie;
+
+import com.google.gson.Gson;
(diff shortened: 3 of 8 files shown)


src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
```diff
@@ -0,0 +1,56 @@
...
+     * @throws IOException
+     */
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json; charset=utf-8");
+
+        Integer id = null;
+
+        if(null != request.getParameter("id")) {
+            id = Integer.parseInt(request.getParameter("id"));
```
henryksloan, line 43: `Integer.parseInt` can throw NumberFormatException, it might be a good idea to catch it with `response.sendError`.

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        if(null == id) {
```
henryksloan, line 46: Are you sure `Integer.parseInt` can return null? I think it returns a primitive `int` on success, and throws NumberFormatException on failure. I think `id` should be a primitive `int`, an error should be shown if 1) getParameter("id") is null, or 2) Integer.parseInt throws NumberFormatException

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
TheGuyWhoCodes: oof. good thought thank you, yeah that would be kinda awkward if someone sent 
```
{
    id: "lol"
}
```

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
henryksloan: Edit: I know the program could theoretically get here if the parameter is null, but the solution I suggested would capture both empty and erroneous Id's.

TheGuyWhoCodes: you're 100% right, thanks!

pom.xml
```diff
@@ -68,6 +68,13 @@
       <artifactId>themoviedbapi</artifactId>
       <version>1.10</version>
     </dependency>
+
```
kirantumkur, line 71: Is it possible to add a comment explaining what this library is doing?

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
```diff
@@ -0,0 +1,68 @@
...
+import info.movito.themoviedbapi.TmdbMovies;
+import info.movito.themoviedbapi.model.MovieDb;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
```
kirantumkur, line 15: Lets remove this, as we get this from the pr, anyways.

pom.xml
```diff
@@ -68,6 +68,13 @@
       <artifactId>themoviedbapi</artifactId>
```
kirantumkur, line 68: Lets add a comment, saying that this library is calling OMDB APIs

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
```diff
@@ -0,0 +1,68 @@
...
+        if(null != request.getParameter("id")) {
+            id = parseInt(request.getParameter("id"));
+            if(null == id) {
+                response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+                return;
+            }
+        } else {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
```
kirantumkur, line 48: Can this be rewritten as?  

 value = request.getParameter("id")
   if value == null {
        response.sendError(HttpServletResponse.SC_BAD_REQUEST);
            return;
   }
 
   id = parseInt(value)
   if id == null {
      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
            return;
   }

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
```diff
@@ -0,0 +1,68 @@
...
+
+        response.getWriter().println(gson.toJsonTree(queryResponse));
+    }
+
+    /**
+     * parseInt method is used to parse an integer from a string (from a url param)
+     * @param urlParam: the url param to convert from string -> integer
+     * @return either an integer if the string is a valid integer, or null if it can't be parsed
+     */
+    private Integer parseInt(String urlParam) {
```
kirantumkur, line 61: rename urlParam to value or something meaningful to this function. A good idea is to reuse what ever Integer.parseInt() uses..

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
```diff
@@ -0,0 +1,68 @@
...
+        try {
+            return Integer.parseInt(urlParam);
+        } catch(NumberFormatException e) {
```
kirantumkur, line 64: What if

src/test/java/com/google/sps/servlets/movie/MovieSearchServletTest.java
```diff
@@ -0,0 +1,90 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/**
```
kirantumkur, line 23: Please remove "author" from the code everywhere. Code is owned and edited by the whole team, and adding "author" like this makes people avoid wanting to code in areas "owned" by others.

src/main/resources/config.properties
```diff
@@ -0,0 +1 @@
+movie_apikey=insert_here
```
kirantumkur, line 1: For now, please add a comment in this file, which explains how the key is generated and shared between team members. We can follow up with a solution later.

.gitignore
```diff
@@ -88,4 +88,6 @@ npm-debug.log
 *.log
 *.stackdump
 scripts/docs.angularjs.org-firebase/functions/content
-/firebase.json
\ No newline at end of file
+/firebase.json
+
+*.properties
```
kirantumkur, line 93: Ideally, this should go into a separate PR.

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
TheGuyWhoCodes: Is rewriting as this fine?

```
        Integer movieId = parseInt(request.getParameter("id"));

        if (movieId == null) {
            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
            return;
        }
```

The parseInt() function handles any null returned by request.getParameter("id") so we can consolidate it into one if block.

src/main/resources/config.properties
TheGuyWhoCodes: Fixed with 974e7117

src/test/java/com/google/sps/servlets/movie/MovieSearchServletTest.java
TheGuyWhoCodes: Fixed with a70f4e0d

pom.xml
TheGuyWhoCodes: Fixed with 631568ce

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
TheGuyWhoCodes: Fixed with a70f4e0d

pom.xml
TheGuyWhoCodes: Fixed with 631568ce

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
TheGuyWhoCodes:
```
        try {
            return Integer.parseInt(urlParam);
        } catch(Exception e) {
            return null;
        }
```

So were you looking for something like this? I looked through `Integer.parseInt()` and I couldn't find any other instances from an error thrown. The method signature is: 
`    public static int parseUnsignedInt(String s, int radix)
                throws NumberFormatException`

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
kirantumkur: Yes. New code makes more sense

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
kirantumkur: Yes, this makes more sense as "all" exceptions are being handled the same way.

henryksloan: https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#parseInt(java.lang.String)
I believe this particular function can only ever throw NumberFormatException. My understanding is that functions have to declare all exceptions they could throw, so I'm pretty sure catching NumberFormatException is more specific, and will never crash. Correct me if I'm wrong.

kirantumkur: Normally, you don't catch "specific" exceptions. You only catch "specific" exceptions, if you have a "specific" action for that type of exception. Or you just catch all exceptions apply the default handler.

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
kirantumkur: Eg: Integer.partInt() is using (string s) as the input. urlParam doesn't make sense, inside of parseInt.

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
henryksloan: Good point, you're right.

.gitignore
TheGuyWhoCodes: Fixed with 0d8a5d1

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
TheGuyWhoCodes: Fixed with cc10931

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
TheGuyWhoCodes: Fixed with 131a2a94

src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java
TheGuyWhoCodes: Fixed with 131a2a94

PR #5, opened 2020-06-22
Add servlet allowing for book searching with the Google Books API

henryksloan:
This servlet parallels the `MovieSearchServlet`. It creates a Books API accessor and creates a query. Since the pagination of the Books API works differently from that of the movies API, we are currently creating pages with 10 results. There is also not an intrinsic results/page maximum from the Google API, so we should probably create our own system of result limiting to unify this.

### DIFF
--- a/src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
+++ b/src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
@@ -40,19 +40,6 @@ public class BookSearchServlet extends HttpServlet {
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
         response.setContentType("application/json; charset=UTF-8");
 
-        final NetHttpTransport HTTP_TRANSPORT;
-        try {
-            HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
-        }
-        catch (GeneralSecurityException e) {
-            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
-            return;
-        }
-
-        Books books = new Books.Builder(HTTP_TRANSPORT, JSON_FACTORY, null)
-                .setApplicationName(APPLICATION_NAME)
-                .build();
-
         int pageNumber;
         try {
             String numStr = request.getParameter("pageNumber");
@@ -69,9 +56,23 @@ public class BookSearchServlet extends HttpServlet {
             return;
         }
 
-       Volumes volumes = books.volumes().list(query)
-               .setMaxResults(10L)
-               .setStartIndex(pageNumber*10L)
+        final NetHttpTransport HTTP_TRANSPORT;
+        try {
+            // Can throw an exception if trusted certificate cannot be established
+            HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+        }
+        catch (GeneralSecurityException e) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+
+        Books books = new Books.Builder(HTTP_TRANSPORT, JSON_FACTORY, null)
+                .setApplicationName(KeyConfig.APPLICATION_NAME)
+                .build();
+
+        Volumes volumes = books.volumes().list(query)
+               .setMaxResults(20L)
+               .setStartIndex(pageNumber*20L)
                .set("country", "US")
                .execute();
 
@@ -91,8 +92,7 @@ public class BookSearchServlet extends HttpServlet {
         // json.put("totalResults", searchResults.getTotalResults());
         // TODO: This doesn't work with the Books API b/c there are practically infinite pages. Maybe limit max pages.
         // json.put("totalPages", searchResults.getTotalPages());
-        // TODO: Client should know this; should it be part of a response?
-        // json.put("page", pageNumber);
+        json.put("page", pageNumber);
 
         return json;
     }
--- a/src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
+++ b/src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
@@ -1,0 +1,99 @@
+package com.google.sps.servlets.book;
+
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.services.books.model.Volumes;
+import com.google.gson.Gson;
+import org.json.simple.JSONObject;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+import com.google.api.services.books.Books;
+
+/**
+ * @author henry
+ * date: 6/19/2020
+ */
+@WebServlet("/books/search")
+public class BookSearchServlet extends HttpServlet {
+
+    private static final String APPLICATION_NAME = "mediaphile";
+    private static JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+
+    private Gson gson = new Gson();
+    private JSONObject json = new JSONObject();
+
+    /**
+     * doGet() handles search queries to Books database.
+     * @param request: a request may have the following query params: query, pageNumber
+     * @param response: a json object returning pagination info and results
+     * @throws IOException:
+     */
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json; charset=UTF-8");
+
+        final NetHttpTransport HTTP_TRANSPORT;
+        try {
+            HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+        }
+        catch (GeneralSecurityException e) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+
+        Books books = new Books.Builder(HTTP_TRANSPORT, JSON_FACTORY, null)
+                .setApplicationName(APPLICATION_NAME)
+                .build();
+
+        int pageNumber;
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,99 @@
...
+import com.google.api.services.books.Books;
+
+/**
+ * @author henry
+ * date: 6/19/2020
+ */
+@WebServlet("/books/search")
+public class BookSearchServlet extends HttpServlet {
+
+    private static final String APPLICATION_NAME = "mediaphile";
```
TheGuyWhoCodes, line 27: can we keep final's like this in the keyconfig or some constants file?

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,99 @@
...
+     * @param response: a json object returning pagination info and results
+     * @throws IOException:
+     */
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json; charset=UTF-8");
+
+        final NetHttpTransport HTTP_TRANSPORT;
+        try {
+            HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
```
TheGuyWhoCodes, line 45: when does this fail? i'm just curious

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,99 @@
...
+        catch (NumberFormatException e) {
+            pageNumber = 0;
+        }
+
+        String query = request.getParameter("query");
+        if (query == null || query.equals("")) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+
```
TheGuyWhoCodes, line 71: should we keep getParameter functions towards the top to keep actual algo and meaningful functions by themselves separately?

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,99 @@
...
+    private JSONObject convertResultsToJson(Volumes searchResults) {
+        JSONObject json = new JSONObject();
+
+        json.put("results", gson.toJsonTree(searchResults.getItems()));
+        // TODO: We should check if there is a parallel for this in Books, but I think it's superfluous to show in general
+        // json.put("totalResults", searchResults.getTotalResults());
+        // TODO: This doesn't work with the Books API b/c there are practically infinite pages. Maybe limit max pages.
+        // json.put("totalPages", searchResults.getTotalPages());
+        // TODO: Client should know this; should it be part of a response?
+        // json.put("page", pageNumber);
```
TheGuyWhoCodes, line 95: i'd just throw this in, doesn't hurt anything

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,99 @@
...
+     * convertResultsToJson converts a Volumes object to json to return to an API request
+     * @param searchResults volumes results from a given query
+     * @return json payload ready to send to user
+     */
+    private JSONObject convertResultsToJson(Volumes searchResults) {
+        JSONObject json = new JSONObject();
+
+        json.put("results", gson.toJsonTree(searchResults.getItems()));
+        // TODO: We should check if there is a parallel for this in Books, but I think it's superfluous to show in general
+        // json.put("totalResults", searchResults.getTotalResults());
```
TheGuyWhoCodes, line 91: what does the getTotalResutls give back? does that just shove all the paginated data into one object?

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
henryksloan: Good point, done.

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
henryksloan: Totally, doing that now.

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
henryksloan: Added a comment. It could potentially fail because .newTrustedTransport calls GoogleUtils.getCertificateTrustStore() which can fail if the certificate can't be received. As far as I understand, this should never happen for our deployment plans, but the exception safety is a good thing.

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
henryksloan: Those commented calls are from the movie search servlet, and don't have a parallel in the Google API. As per our discussions about this, I'm currently limiting results per page to 20 (that's what the movies API seems to return per page). We won't have a problem returning too much metadata, but we should unify the API return values once we start integrating it with Datastore.

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
henryksloan: I will for now to match your servlet, but we'll work out a common output format later on.

TheGuyWhoCodes:
i think this pr looks good, just needs test

TheGuyWhoCodes:
LGTM, approved

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,95 @@
...
+        String query = request.getParameter("query");
+        if (query == null || query.equals("")) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+
+        final NetHttpTransport HTTP_TRANSPORT;
+        try {
+            // Can throw an exception if trusted certificate cannot be established
+            HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
```
kirantumkur, line 58: @Henry, didnt we discuss camelcase for all variable names?

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json; charset=UTF-8");
+
+        int pageNumber;
+        try {
+            String numStr = request.getParameter("pageNumber");
+            if (numStr == null) numStr = "0";
+            pageNumber = Integer.parseInt(numStr);
```
kirantumkur, line 43: Look at https://github.com/kirantumkur/mediaphile/blob/8187d74e0e0a0a2e32aed3d71de372d6d2d2f948/src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java#L55

Move "parseInt" method to a common location, and reuse that function. 

For Example, Move "parseInt" method to src/main/java/com/google/sps/servlets/util/Utils.java

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,95 @@
...
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+
+        final NetHttpTransport HTTP_TRANSPORT;
+        try {
+            // Can throw an exception if trusted certificate cannot be established
+            HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+        }
+        catch (GeneralSecurityException e) {
```
kirantumkur, line 60: Use a "specific" type of exception, only if you are having a specifc error handler. If your error handler is generic, use "catch all\default"

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,95 @@
...
+            return;
+        }
+
+        Books books = new Books.Builder(HTTP_TRANSPORT, JSON_FACTORY, null)
+                .setApplicationName(KeyConfig.APPLICATION_NAME)
+                .build();
+
+        Volumes volumes = books.volumes().list(query)
+               .setMaxResults(20L)
+               .setStartIndex(pageNumber*20L)
```
kirantumkur, line 71: Define a constant for NumLines at the top of this file and set that to 20L.

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -0,0 +1,95 @@
+package com.google.sps.servlets.book;
```
kirantumkur, line 1: Add tests for this file.

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
henryksloan: Yes, I'll fix that today and update this PR. This PR came out before we discussed that, and I've been working on other features as this was being reviewed.

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
henryksloan: Good point, we'll discuss common functions that can be moved to a utilities file.

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
henryksloan: You're right, same reason as the CamelCase comment above. I'll fix both today.

src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
henryksloan: I intended to push tests for both book searching and details as a separate PR. Would you prefer that I do that, or add them here?

kirantumkur: Please add tests in this PR itself. Let us always try to check in logic code and test code in the same PR.

henryksloan: Sounds good, I'll work on that today.

PR #6, opened 2020-06-24
Add book details servlet with error safety and proper response

henryksloan:
Added a servlet that requires an `id` parameter, builds a Books API accessor, dispatches a GET request for the particular item, and returns a JSON encoding of the response.

### DIFF
--- a/src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
+++ b/src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
@@ -19,7 +19,7 @@ import java.security.GeneralSecurityException;
 @WebServlet("/books/details")
 public class BookDetailsServlet extends HttpServlet {
 
-    private static JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+    private static final JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
 
     private Gson gson = new Gson();
 
@@ -41,17 +41,17 @@ public class BookDetailsServlet extends HttpServlet {
             return;
         }
 
-        final NetHttpTransport HTTP_TRANSPORT;
+        final NetHttpTransport httpTransport;
         try {
             // Can throw an exception if trusted certificate cannot be established
-            HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+            httpTransport = GoogleNetHttpTransport.newTrustedTransport();
         }
         catch (Exception e) {
             response.sendError(HttpServletResponse.SC_BAD_REQUEST);
             return;
         }
 
-        Books books = new Books.Builder(HTTP_TRANSPORT, JSON_FACTORY, null)
+        Books books = new Books.Builder(httpTransport, jsonFactory, null)
                 .setApplicationName(KeyConfig.APPLICATION_NAME)
                 .build();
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
```diff
@@ -0,0 +1,62 @@
...
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json; charset=utf-8");
+
+        String id = request.getParameter("id");
+        if (id == null || id.equals("")) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+
+        final NetHttpTransport HTTP_TRANSPORT;
```
kirantumkur, line 44: use camelCase for variable names

src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
```diff
@@ -0,0 +1,62 @@
...
+        try {
+            // Can throw an exception if trusted certificate cannot be established
+            HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+        }
+        catch (GeneralSecurityException e) {
```
kirantumkur, line 49: Switch to catch all exceptions, as there is no specific action being taken for this exception.

src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+@WebServlet("/books/details")
+public class BookDetailsServlet extends HttpServlet {
+
+    private static JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
```
kirantumkur, line 22: please use camelCase for variable names

src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
henryksloan: Got it, fixed

src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
henryksloan: The Google Java style guide recommends this for constants (https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names), so we've all been using it for final/static final objects. We can change it here and elsewhere if you think the alternative is more appropriate.

kirantumkur: the value of HTTP_TRANSPORT is being changed on line number 47. It is not being used as a constant.

kirantumkur: From the style guide, none of these are changing after compile time. If they change after compile\build time, they are not constants. 


// Constants
static final int NUMBER = 5;
static final ImmutableList<String> NAMES = ImmutableList.of("Ed", "Ann");
static final ImmutableMap<String, Integer> AGES = ImmutableMap.of("Ed", 35, "Ann", 32);
static final Joiner COMMA_JOINER = Joiner.on(','); // because Joiner is immutable
static final SomeMutableType[] EMPTY_ARRAY = {};
enum SomeEnum { ENUM_CONSTANT }

henryksloan: Oh okay, that totally makes sense, thanks. I was thinking of C++ where constants need their values at compile time.

src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
henryksloan: Fixed in 8062ec8

PR #7, opened 2020-06-26
Dummy frontend

XLuna24:
Add search bar, and Add title of movies to front end

### DIFF
--- a/src/main/webapp/home.css
+++ b/src/main/webapp/home.css
@@ -1,0 +1,64 @@
+html, body {
+    margin: 0%;
+    padding: 0%;
+}
+.topnav {
+    overflow: auto;
+    background-color: #f8f8f8;
+    box-shadow: 0px 3px 5px #a0a0a0;
+}
+
+.topnav h1 {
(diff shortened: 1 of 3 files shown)


TheGuyWhoCodes:
looks fine for a quick proof of concept. in the future it would be nice for a search result to have a title, date, movie description and possibly hot adds for queue / watched list

src/main/webapp/home.css
```diff
@@ -0,0 +1,64 @@
+html, body {
+    margin: 0%;
+    padding: 0%;
```
TheGuyWhoCodes, line 3: instead of doing these normalizes, we should instead using something like normalize.css or bootstrap (without features) to normalize our pages.

kirantumkur: @TheGuyWhoCodes , Can you elaborate on this comment and provide instructions on how to do it?

XLuna24: I will add this file later on. right now its just to show the basic list of movies

PR #9, opened 2020-06-29
Add a servlet to retrieve user information from the datastore based on authentication

henryksloan:
UserServlet defines the API endpoint /user, accepting a GET request with a required `id` parameter. The request throws an error if the `id` parameter is missing or empty, or if no matching user is found in the Datastore. The returned email field is equal to "" unless the requester is both authenticated and has a matching `id` to the query (i.e. users can only see their own emails).

UserServletTest provides several tests of this servlet. Leveraging the appengine-testing library (added to `pom.xml`, along with a co-dependency), it emulates various states of the Datastore and UserService. The tests ensure:

1. That query for a nonexistent user fails properly
2. That unauthenticated or other-user access to an existing user returns all information except for the email
3. That a user querying their own `id` returns all information, including email
4. That empty and missing `id` query parameters cause a proper error

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,26 +59,28 @@
 
     <dependency>
       <groupId>com.google.appengine</groupId>
-      <artifactId>appengine-api-1.0-sdk</artifactId>
-      <version>1.9.59</version>
-    </dependency>
-
-    <!-- https://mvnrepository.com/artifact/com.google.appengine/appengine-testing -->
-    <!-- Provides helper classes to simulate UserService and Datastore states -->
-    <dependency>
-      <groupId>com.google.appengine</groupId>
       <artifactId>appengine-testing</artifactId>
-      <version>1.9.80</version>
+      <version>1.9.59</version>
       <scope>test</scope>
     </dependency>
 
-    <!-- https://mvnrepository.com/artifact/com.google.appengine/appengine-api-stubs -->
-    <!-- appengine-testing depends on this -->
     <dependency>
       <groupId>com.google.appengine</groupId>
       <artifactId>appengine-api-stubs</artifactId>
-      <version>1.9.80</version>
+      <version>1.9.59</version>
       <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-api-labs</artifactId>
+      <version>1.9.59</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-tools-sdk</artifactId>
+      <version>1.9.59</version>
     </dependency>
 
     <!--    This is a wrapper for The Movie DB API. Here is the github: https://github.com/holgerbrandl/themoviedbapi-->
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,7 @@
     </dependency>
 
     <!-- https://mvnrepository.com/artifact/com.google.appengine/appengine-testing -->
+    <!-- Provides helper classes to simulate UserService and Datastore states -->
     <dependency>
       <groupId>com.google.appengine</groupId>
       <artifactId>appengine-testing</artifactId>
@@ -72,6 +73,7 @@
     </dependency>
 
     <!-- https://mvnrepository.com/artifact/com.google.appengine/appengine-api-stubs -->
+    <!-- appengine-testing depends on this -->
     <dependency>
       <groupId>com.google.appengine</groupId>
       <artifactId>appengine-api-stubs</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,24 @@
       <artifactId>appengine-api-1.0-sdk</artifactId>
       <version>1.9.59</version>
     </dependency>
-<!--    This is a wrapper for The Movie DB API. Here is the github: https://github.com/holgerbrandl/themoviedbapi-->
+
+    <!-- https://mvnrepository.com/artifact/com.google.appengine/appengine-testing -->
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-testing</artifactId>
+      <version>1.9.80</version>
+      <scope>test</scope>
+    </dependency>
+
+    <!-- https://mvnrepository.com/artifact/com.google.appengine/appengine-api-stubs -->
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-api-stubs</artifactId>
+      <version>1.9.80</version>
+      <scope>test</scope>
+    </dependency>
+
+    <!--    This is a wrapper for The Movie DB API. Here is the github: https://github.com/holgerbrandl/themoviedbapi-->
 <!--    Here is the API Docs: https://developers.themoviedb.org/-->
     <dependency>
       <groupId>info.movito</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -59,6 +59,12 @@
 
     <dependency>
       <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-api-1.0-sdk</artifactId>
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/sps/servlets/user/UserServlet.java
```diff
@@ -0,0 +1,101 @@
...
+            this.username = username;
+            this.email = email;
+            this.profilePicUrl = profilePicUrl;
+        }
+    }
+
+    // Search Datastore for an entry for the user
+    public static Entity checkDatastore(String id) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        Query query = new Query("User")
```
TheGuyWhoCodes, line 45: in the future we should probably refactor this to use that objectify library we talked about

pom.xml
```diff
@@ -62,7 +62,26 @@
...
+    <!-- https://mvnrepository.com/artifact/com.google.appengine/appengine-testing -->
+    <!-- Provides helper classes to simulate UserService and Datastore states -->
+    <dependency>
+      <groupId>com.google.appengine</groupId>
+      <artifactId>appengine-testing</artifactId>
+      <version>1.9.80</version>
+      <scope>test</scope>
+    </dependency>
+
+    <!-- https://mvnrepository.com/artifact/com.google.appengine/appengine-api-stubs -->
```
TheGuyWhoCodes, line 75: When importing these tests we need to make to grab all of the testing libraries out that are needed for app engine: ie

```
    <dependency>
      <groupId>com.google.appengine</groupId>
      <artifactId>appengine-testing</artifactId>
      <version>1.9.59</version>
      <scope>test</scope>
    </dependency>

    <dependency>
      <groupId>com.google.appengine</groupId>
      <artifactId>appengine-api-stubs</artifactId>
      <version>1.9.59</version>
      <scope>test</scope>
    </dependency>

    <dependency>
      <groupId>com.google.appengine</groupId>
      <artifactId>appengine-api-labs</artifactId>
      <version>1.9.59</version>
    </dependency>

    <dependency>
      <groupId>com.google.appengine</groupId>
      <artifactId>appengine-tools-sdk</artifactId>
      <version>1.9.59</version>
    </dependency>
```

TheGuyWhoCodes: and also don't mismatch version numbers, we can either stay at the 1.9.59 that we started with or go up to the version you're using

src/main/java/com/google/sps/servlets/user/UserServlet.java
```diff
@@ -0,0 +1,101 @@
...
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json; charset=utf-8");
+
+        UserService userService = UserServiceFactory.getUserService();
+        User user = userService.getCurrentUser();
+
+        String id = request.getParameter("id");
+        if (id == null || id.equals("")) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+            return;
```
TheGuyWhoCodes, line 86: this can be refactored into a function and possibly used to have an actual body sent out with it instead of just empty, ie

```
    private void sendInvalidGetResponse(HttpServletResponse response) {
        response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
        ReturnObject newReturn = new ReturnObject("error");
    }
```

src/main/java/com/google/sps/servlets/user/UserServlet.java
```diff
@@ -0,0 +1,101 @@
...
+            (String) entity.getProperty("username"),
+            (String) entity.getProperty("email"),
+            (String) entity.getProperty("profilePicUrl")
+        );
+    }
+
+    /**
+     * doGet() returns details of the particular user with the given id
+     * Returns error 400 if no id is provided
+     * Returns error 500 if the user cannot be found
```
TheGuyWhoCodes, line 71: 500 i don't think would be the right response code for this. since everything is working, the only issue is we couldn't find the stored user.

pom.xml
kirantumkur: Please resolve comments after they are addressed.

src/main/java/com/google/sps/servlets/user/UserServlet.java
kirantumkur: Look at

mediaphile/src/main/java/com/google/sps/servlets/movie/MovieDetailsServlet.java

Line 55 in 8187d74

 private Integer parseInt(String s) { 
Move "parseInt" method to a common location, and reuse that function.

For Example, Move "parseInt" method to src/main/java/com/google/sps/servlets/util/Utils.java

src/main/java/com/google/sps/servlets/user/UserServlet.java
```diff
@@ -0,0 +1,101 @@
...
+            this.email = email;
+            this.profilePicUrl = profilePicUrl;
+        }
+    }
+
+    // Search Datastore for an entry for the user
+    public static Entity checkDatastore(String id) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        Query query = new Query("User")
+                .setFilter(new Query.FilterPredicate("id", Query.FilterOperator.EQUAL, id));
```
kirantumkur, line 46: consider getting by "email" instead of id. As email is going to be unique. Id may run into issues.

src/main/java/com/google/sps/servlets/user/UserServlet.java
```diff
@@ -0,0 +1,101 @@
...
+     * @throws IOException
+     */
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json; charset=utf-8");
+
+        UserService userService = UserServiceFactory.getUserService();
+        User user = userService.getCurrentUser();
+
+        String id = request.getParameter("id");
```
kirantumkur, line 83: use email.. id is going to run into issues with uniqueness.

src/main/java/com/google/sps/servlets/user/UserServlet.java
TheGuyWhoCodes: this sounds like a great idea

henryksloan: The three of us discussed this earlier today; we're working on a common solution for error pages that will let us flesh them out later on. I think for now, one of us will create a common error-return utility (I think Chris has something like that in a working branch) and we will use it in places like this.

src/main/java/com/google/sps/servlets/user/UserServlet.java
kirantumkur: How about 404, NotFound?

https://www.restapitutorial.com/httpstatuscodes.html

src/main/java/com/google/sps/servlets/user/UserServlet.java
henryksloan: https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/users/User#getuserid
userId comes from the userService, and is defined as uniquely identifying. Email will generally be unique as well, but is there a reason we should avoid using the generic identifier?

src/main/java/com/google/sps/servlets/user/UserServlet.java
henryksloan: Fixed using 404

src/main/java/com/google/sps/servlets/user/UserServlet.java
henryksloan: Good point. Once this PR is merged, I'll rebase LoginServlet and refactor UserServlet and LoginServlet with Objectify.

src/main/java/com/google/sps/servlets/user/UserServlet.java
kirantumkur: User id between multiple services might clash. Emails are universally unique.

PR #11, opened 2020-06-30
Implements Queue + Previously Watched POST and GET Endpoints

TheGuyWhoCodes:
## Majors
 - Implements Queue + Previously Watched POST and GET Endpoints
 - Implements tests for endpoints
 - Adds helper Input Streamer for tests

## Minors
 - Adds google appengine testing dependencies

### DIFF
--- a/src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
+++ b/src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
@@ -39,12 +39,12 @@ public class EntityListServlet extends HttpServlet {
         String queueType = request.getParameter("entityType");
 
         if(queueType == null) {
-            sendInvalidGetResponse(response);
+            setInvalidGetResponse(response);
             return;
         }
 
         if(!isCorrectListType(queueType)) {
-            sendInvalidGetResponse(response);
+            setInvalidGetResponse(response);
             return;
         }
 
@@ -161,7 +161,7 @@ public class EntityListServlet extends HttpServlet {
      * Sets status to 400 if we can't parse input
      * @param response: response to add status to
      */
-    private void sendInvalidGetResponse(HttpServletResponse response) {
+    private void setInvalidGetResponse(HttpServletResponse response) {
         response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
     }
 
--- a/src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
+++ b/src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
@@ -72,11 +72,6 @@ public class EntityListServlet extends HttpServlet {
                 return;
             }
 
-            // couldn't parse type correctly, will return a failed attempt
-            if(entityDb == null) {
-                sendInvalidPostResponse(response, newResponse);
-                return;
-            }
             try {
                 // Entry being saved to the datastore instance
                 ofy().save().entity(entityDb).now();
@@ -132,14 +127,14 @@ public class EntityListServlet extends HttpServlet {
      * @return: null if invalid entity type, an EntityDbQueue type object if it can
      * @throws JsonProcessingException: if it can't parse out data.
      */
-    private EntityDbQueue decideDbType(String body) throws JsonProcessingException {
+    private EntityDbQueue decideDbType(String body) throws JsonProcessingException, NoSuchFieldException {
         EntityDbQueue entity = mapper.readValue(body, EntityDbQueue.class);
         if(entity.getEntityType().equals(typeViewed)) {
             return mapper.readValue(body, WatchedQueueObject.class);
         } else if(entity.getEntityType().equals(typeQueue)) {
             return mapper.readValue(body, WantToWatchQueueObject.class);
         } else {
-            return null;
+            throw new NoSuchFieldException();
         }
     }
 
--- a/src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
+++ b/src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
@@ -29,8 +29,8 @@ public class EntityListServlet extends HttpServlet {
     private final Gson gson = new GsonBuilder().serializeNulls().create();
     private final ObjectMapper mapper = new ObjectMapper();
 
-    private final String TYPE_QUEUE = "queue";
-    private final String TYPE_VIEWED = "viewed";
+    private final String typeQueue = "queue";
+    private final String typeViewed = "viewed";
 
     @Override
     protected void doGet(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
@@ -106,7 +106,7 @@ public class EntityListServlet extends HttpServlet {
      * @return: a list of abstracted type EntityDbQueue
      */
     private List<? extends EntityDbQueue> getQueueWithType(String type, String userID) {
-        if(type.equals(TYPE_QUEUE)) {
+        if(type.equals(typeQueue)) {
             return ofy().load().type(WantToWatchQueueObject.class).filter("userID", userID).list();
         } else {
             return ofy().load().type(WatchedQueueObject.class).filter("userID", userID).list();
@@ -119,7 +119,7 @@ public class EntityListServlet extends HttpServlet {
      * @return: true / false if the type is valid
      */
     private boolean isCorrectListType(String queueType) {
-        return queueType.equals(TYPE_QUEUE) || queueType.equals(TYPE_VIEWED);
+        return queueType.equals(typeQueue) || queueType.equals(typeViewed);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
```diff
@@ -0,0 +1,171 @@
...
+
+@WebServlet("/list/entity")
+public class EntityListServlet extends HttpServlet {
+
+    private final UserService userService = UserServiceFactory.getUserService();
+    private final Gson gson = new GsonBuilder().serializeNulls().create();
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    private final String TYPE_QUEUE = "queue";
+    private final String TYPE_VIEWED = "viewed";
```
kirantumkur, line 33: Do you think we can use enum's here?

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json; charset=utf-8");
+        QueueResponse newResponse = new QueueResponse();
+        String body = parseBody(request);
+        EntityDbQueue entityDb = null;
+        User user = userService.getCurrentUser();
+        if(userService.isUserLoggedIn()) {
```
kirantumkur, line 62: Is the logged in user here, the service account running the server or the actual customer using the system?

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
kirantumkur, line 33: camelCase variable names

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            // couldn't parse type correctly, will return a failed attempt
+            if(entityDb == null) {
+                sendInvalidPostResponse(response, newResponse);
+                return;
+            }
+
+            ofy().save().entity(entityDb).now();
+
```
kirantumkur, line 82: does this throw? how do we know it always succeeds? Also, add a comment saying that the entity is being saved.

src/test/java/com/google/sps/servlets/DelegatingServletInputStream.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public boolean isReady() {
+        return false;
+    }
+
+    @Override
+    public void setReadListener(ReadListener readListener) {
+
+    }
```
kirantumkur, line 56: why is this empty?

kirantumkur: if this doesnt need implementation right now, use NotImplementedException() or something similar to that.

src/test/java/com/google/sps/servlets/DelegatingServletInputStream.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.servlet.ReadListener;
+import javax.servlet.ServletInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Helper function to help test null data
+ */
+public class DelegatingServletInputStream extends ServletInputStream {
+
```
kirantumkur, line 14: Rename to start with Test or Fake - as this is test data and this shouldn't be used in regular code.

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
```diff
@@ -0,0 +1,171 @@
...
+        String userID = request.getParameter("userID");
+        String queueType = request.getParameter("entityType");
+
+        if(queueType == null) {
+            sendInvalidGetResponse(response);
+            return;
+        }
+
+        if(!isCorrectListType(queueType)) {
+            sendInvalidGetResponse(response);
```
kirantumkur, line 47: nit: rename sendInvalidGetResponse to setInvalidGetResponse

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * A helper function to get entity data from database
+     * @param type: what list type should it look for (queue / watched)
+     * @param userID: userID to look for
+     * @return: a list of abstracted type EntityDbQueue
+     */
+    private List<? extends EntityDbQueue> getQueueWithType(String type, String userID) {
+        if(type.equals(TYPE_QUEUE)) {
+            return ofy().load().type(WantToWatchQueueObject.class).filter("userID", userID).list();
```
kirantumkur, line 110: do these commands throw? how do you know that they will not fail?

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
```diff
@@ -0,0 +1,171 @@
...
+        } else {
+            return ofy().load().type(WatchedQueueObject.class).filter("userID", userID).list();
+        }
```
kirantumkur, line 113: do these commands throw? how do you know that they will not fail?

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
TheGuyWhoCodes: According to the docs here: https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/users/UserService#isuserloggedin it is the actual user logged in.

src/test/java/com/google/sps/servlets/DelegatingServletInputStream.java
TheGuyWhoCodes: Because I'm not abstracting out the class, Java wants all methods to be overridden. I will add in that exception.

src/test/java/com/google/sps/servlets/DelegatingServletInputStream.java
TheGuyWhoCodes: Good idea!

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
TheGuyWhoCodes: Fix with 0b45504

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
TheGuyWhoCodes: After some investigation, it can throw a NullPointerException if the entity is null. Adding a try catch will add an extra layer of safety.

TheGuyWhoCodes: Fix with 
6d412df

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
TheGuyWhoCodes: From what I have found, these do not. These will be safely loaded and return null if nothing comes back.

src/test/java/com/google/sps/servlets/DelegatingServletInputStream.java
TheGuyWhoCodes: Fixed with 80e5e12

src/test/java/com/google/sps/servlets/DelegatingServletInputStream.java
TheGuyWhoCodes: Fixed with 80e5e12

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
TheGuyWhoCodes: Fixed with 
aa84692

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
kirantumkur: Lets chat about this. I dont understand this 

                entityDb = decideDbType(body);
                if(!currentUserID.equals(entityDb.getUserID())) {
                    sendInvalidPostResponse(response, newResponse);
                    return;
                }

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
```diff
@@ -0,0 +1,175 @@
...
+     * @throws JsonProcessingException: if it can't parse out data.
+     */
+    private EntityDbQueue decideDbType(String body) throws JsonProcessingException {
+        EntityDbQueue entity = mapper.readValue(body, EntityDbQueue.class);
+        if(entity.getEntityType().equals(typeViewed)) {
+            return mapper.readValue(body, WatchedQueueObject.class);
+        } else if(entity.getEntityType().equals(typeQueue)) {
+            return mapper.readValue(body, WantToWatchQueueObject.class);
+        } else {
+            return null;
```
kirantumkur, line 142: Instead of a null, is it not better to throw an exception and catch it?

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
TheGuyWhoCodes: this statement checks to make sure that the user logged in is adding to the correct list. Ie if a userID of 1234 tries to add to userID 12345, it won't work

We can also change this so the user doesn't post the userid in the body, then simply get the current userID from the UserService.

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
TheGuyWhoCodes: sounds like a good idea!

TheGuyWhoCodes: Just fixed with 21d6cdb

PR #12, opened 2020-06-30
Book display

XLuna24:
Displays the list of books or movies according to users search. It Display's both Books and Movies in small boxes with a title and description.

### DIFF
--- a/src/main/webapp/home.js
+++ b/src/main/webapp/home.js
@@ -1,3 +1,11 @@
+function handle(event) {
+    if(event.keyCode === 13) {
+        console.log(event.keyCode);
+        event.preventDefault();
+        loadContent();
+    }
+}
+
 function loadContent() {
     document.getElementById("movie-tab").style.display = "none"
     document.getElementById("book-tab").style.display = "none"
--- a/src/main/webapp/home.js
+++ b/src/main/webapp/home.js
@@ -1,7 +1,28 @@
+function loadContent() {
+    document.getElementById("movie-tab").style.display = "none";
+    document.getElementById("book-tab").style.display = "none";
+    const option = document.getElementById('selection').value;
+
+    if(option === "movie") {
+        document.getElementById("movie-tab").style.display = "block";
+        loadMovies();
+    } else if(option === "book"){
+        document.getElementById("book-tab").style.display = "block"
+        loadBooks();
+    } else {
+        document.getElementById("movie-tab").style.display = "block"
+        loadMovies();
+
+        document.getElementById("book-tab").style.display = "block";
+        loadBooks();
+    }
+}
+
 function loadMovies() {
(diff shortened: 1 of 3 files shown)


src/main/webapp/home.js
```diff
@@ -1,17 +1,75 @@
...
             contentList.appendChild(createContentElement(info));
         })
     });
 }
 
-//Creates an li to display the title of the movies
+function loadBooks() {
+    const search = document.getElementById("search").value;
+
+    fetch('/books/search?query=' + search + "&pageNumber=1")
```
henryksloan, line 42: Book pages actually start at zero. For now, change this to zero so it shows the right results. It definitely makes more sense to query pages as they're shown to the user (i.e. 1 is first), so I'll change my code once we get a chance to talk about it.

src/main/webapp/home.js
```diff
@@ -1,17 +1,75 @@
...
+    const description = document.createElement('div');
+    description.className = 'description-box';
+    description.innerText = info.description ? info.description : info.overview;
+
+    contentElement.appendChild(title);
+    contentElement.appendChild(description);
     return contentElement;
+}
+
+//Display's wether the movies or books list appears
```
henryksloan, line 69: Displays whether

src/main/webapp/index.html
```diff
@@ -12,18 +12,24 @@
           <h1>Mediaphile</h1>
           <div class="search-container">
                   <input type="text" placeholder="Search.." id="search">
-                  <button class="button" onclick="loadMovies()">Submit</button>
+                  <button class="button" onclick="loadContent()">Submit</button>
+                  <label>in</label>
+                  <select id="selection" class="button">
+                      <option value="movie">Movies</option>
+                      <option value="book">Books</option>
+                      <option value="all">All</option>
```
henryksloan, line 20: I imagine "All" would be on top/default, but don't change this for now. We'll all discuss that as we think about the search journey.

TheGuyWhoCodes: using all will bring up some challenges for standardizing how we interpret search results ie a book entity and a movie entity from a search result are different from each other.

XLuna24: has been switched to movies/books and made default so later on we can add in the user search

src/main/webapp/home.js
```diff
@@ -1,17 +1,83 @@
...
     return contentElement;
+}
+
+//Display's whether the movies or books list appears
+function displayContent(content) {
+    document.getElementById("movie-list").style.display = "none";
+    document.getElementById("book-list").style.display = "none";
+
+    document.getElementById(content).style.display = "block";
 }
```
kirantumkur, line 83: This code has a lot of "const" variables which are being reassigned values. Please verify that "const" is being used correctly here.

PR #14, opened 2020-07-06
Implement servlet to provide login information and links, and servlet to register new users to the Datastore

henryksloan:
LoginStatus is a POJO providing information about the login status of a user, whether or not they are logged in. It provides user's boolean login state, their userId (if they are logged in), and a login/logout URL provided by userService. The ID will be used to produce the links to the profile page, etc. The URL field is a link to the Google login provider, allowing a user to log in or out of the website through a Google authenticator (or a dummy login page on the local dev server).

LoginServlet provides the endpoint GET /login. It returns a LoginStatus object based on the user state according to userService. This service provides the login/logout URL with the appropriate redirects. Logging out redirects to the homepage "/", while logging in redirects to /login/callback (see below), which then redirects to "/".

LoginCallbackServlet provides the endpoint GET /login/callback. If no user is logged in at the time of the GET request according to the userService, then it throws error 400. Otherwise, it checks the Datastore for a user with the ID of the logged in user (i.e. it checks if the user is registered in the Datastore). If they are registered, then nothing is done, and the user is redirected without side effect. If the user is not registered, their information is procured from userService and pushed to the Datastore.

LoginServletTest ensures that login/logout URLs are always returned, that logged out users are given {loggedIn: false, id: ""}, and that logged in users are given {loggedIn: true} and the correct user ID.

LoginCallbackTest ensures that LoginCallback throws an error when the user is not logged in, that new users are registered and redirected, and that already registered users remain unchanged in the Datastore.

### DIFF
--- a/src/main/java/com/google/sps/servlets/user/LoginCallbackServlet.java
+++ b/src/main/java/com/google/sps/servlets/user/LoginCallbackServlet.java
@@ -12,7 +12,7 @@ import com.google.appengine.api.users.User;
 
 import static com.googlecode.objectify.ObjectifyService.ofy;
 
-/** Servlet that handles user logins. */
+/** Servlet that registers new users after login. */
 @WebServlet("/login/callback")
 public class LoginCallbackServlet extends HttpServlet {
 
--- a/src/main/java/com/google/sps/servlets/user/LoginCallbackServlet.java
+++ b/src/main/java/com/google/sps/servlets/user/LoginCallbackServlet.java
@@ -22,9 +22,10 @@ public class LoginCallbackServlet extends HttpServlet {
     }
 
     /**
-     * doGet() returns details of the particular volume with the given id
-     * Returns error 400 if no id is provided
-     * Returns error 500 if no movie is returned by the API
+     * doGet() attempts to register a newly logged in user to the Datastore
+     * Returns error 400 if the user is not logged in
+     * Otherwise redirects to the homepage
+     * If the logged in user is not registered in the Datastore, they are added
      * @param request: expects id parameter
      * @param response: returns a Volume object
      * @throws IOException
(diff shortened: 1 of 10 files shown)


TheGuyWhoCodes:
looks good, could the API endpoint for login be a little more descriptive, maybe `/login/status`?

henryksloan:
@TheGuyWhoCodes Great idea, done

src/main/java/com/google/sps/servlets/user/LoginCallbackServlet.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.appengine.api.users.User;
+
+import static com.googlecode.objectify.ObjectifyService.ofy;
+
+/** Servlet that registers new users after login. */
+@WebServlet("/login/callback")
+public class LoginCallbackServlet extends HttpServlet {
```
kirantumkur, line 17: Rename to LoginRegistrationServlet.

src/main/java/com/google/sps/servlets/user/LoginServlet.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.IOException;
+
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.appengine.api.users.User;
+import com.google.gson.Gson;
+
+/** Servlet that handles user logins. */
+@WebServlet("/login/status")
+public class LoginServlet extends HttpServlet {
```
kirantumkur, line 16: Rename to LoginStatusServelet

kirantumkur: Update relevant test names too

src/main/java/com/google/sps/servlets/user/LoginCallbackServlet.java
kirantumkur: Update relevant test names too

src/main/java/com/google/sps/servlets/user/LoginServlet.java
henryksloan: Good point, both done (same for registration).

PR #15, opened 2020-07-07
Add Delete endpoint to Lists

TheGuyWhoCodes:
## Majors
 - Add `DELETE` endpoint to delete entities from either watched / queue list.
 - Modifies EntityDb entry to correctly support IDs

## Minors
 - Adds tests for `DELETE` endpoint


src/main/java/com/google/sps/servlets/queue/DeleteEntityFromListServlet.java
```diff
@@ -0,0 +1,77 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+import static com.googlecode.objectify.ObjectifyService.ofy;
+
+/**
+ * Used to delete an element from list (watched / queue)
+ */
+@WebServlet("/list/delete")
+public class DeleteEntityFromListServlet extends HttpServlet {
```
henryksloan, line 26: Shouldn't this be implemented as a doDelete method in the EntityListServlet? That would make EntityListServlet very code-heavy, but I think the best thing to do would be to move the HTTP helper functions like sendNotLoggedIn to a new com.google.sps.util.HttpUtils class, and auxiliary functions like getQueueWithType perhaps to com.google.sps.util.ListUtils (or they could stay in EntityListServlet if it makes it clearer, but that class is a bit too big).

src/main/java/com/google/sps/servlets/queue/DeleteEntityFromListServlet.java
```diff
@@ -0,0 +1,77 @@
...
+            // nothing was found to delete
+            if(Iterables.size(allKeys) == 0) {
+                response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+                return;
+            }
+            // delete from db, will delete ANY instance of that id from the user
+            // ie. if somehow multiple of the same movie / book get saved in the list
+            ofy().delete().keys(allKeys).now();
+        } else {
+            System.out.println("6");
```
henryksloan, line 72: Remove debug code

src/main/java/com/google/sps/servlets/queue/DeleteEntityFromListServlet.java
```diff
@@ -0,0 +1,77 @@
...
+        String listType = request.getParameter("listType");
+
+        // check to make sure all needed fields are there.
+        if(id == null || listType == null) {
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            return;
+        }
+
+        User user = userService.getCurrentUser();
+        if(userService.isUserLoggedIn()) {
```
henryksloan, line 46: Probably change this to
```
if (!userService.isUserLoggedIn()) {
    response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
    return;
}
```
So the majority of the code isn't inside the if block.

src/test/java/com/google/sps/servlets/queue/EntityListServletTest.java
```diff
@@ -166,17 +166,72 @@ public void invalidPostEntity() throws IOException, ServletException {
         verify(response, times(1)).setStatus(400);
     }
 
+
+    @Test
+    public void deleteFromDbWatched() throws IOException, ServletException {
```
henryksloan, line 171: So far we've had one test file for each servlet, so I think we should either move the deletion behavior to EntityListServlet.doDelete(), or create a separate test class (the former is preferable as per my other comment).

src/main/java/com/google/sps/servlets/queue/DeleteEntityFromListServlet.java
TheGuyWhoCodes: Ok, so i'm going to delete the class and move the doDelete over to EntityListServlet. Then, I'll move out some of those returns over to so the HttpUtils class.

src/main/java/com/google/sps/servlets/queue/DeleteEntityFromListServlet.java
TheGuyWhoCodes: oops thank you. for some reason my intellij debugger hasn't been hitting any breakpoints.

src/main/java/com/google/sps/servlets/queue/DeleteEntityFromListServlet.java
TheGuyWhoCodes: good idea!

src/test/java/com/google/sps/servlets/queue/EntityListServletTest.java
TheGuyWhoCodes: Lets keep this in here since i'll be moving the deletion behavior to EntityListServlet

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
```diff
@@ -88,14 +89,50 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
...
-     * @return: Stringified JSON request
-     * @throws IOException: if the body can't be parsed
+     *  the delete endpoint is used to delete an entity from either a "watched" or "queued" list
+     *  needs an "id" representing the entity id, and a "list type" being either queued or watched
      */
-    private String parseBody(HttpServletRequest request) throws IOException {
-        return request.getReader().lines()
-                .reduce("", (accumulator, actual) -> accumulator + actual);
+    @Override
+    protected void doDelete(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
```
kirantumkur, line 96: Please break apart doDelete into modular methods.

src/main/java/com/google/sps/util/Utils.java
```diff
@@ -1,11 +1,38 @@
...
+    /**
+     * Helper function to parse out body data from requests
+     * @param request: the request to parse out from
+     * @return: Stringified JSON request
+     * @throws IOException : if the body can't be parsed
+     */
+    public static String parseBody(HttpServletRequest request) throws IOException {
+        return request.getReader().lines()
+                .reduce("", (accumulator, actual) -> accumulator + actual);
+    }
```
kirantumkur, line 28: Sorry, I didnt understand what this method is doing. Also, how are inputs accumulator and actual being inputted?

src/test/java/com/google/sps/servlets/queue/EntityListServletTest.java
```diff
@@ -35,7 +35,7 @@
     public void initialize() {
         new ContextListener().initDbObjects();
         Map<String, Object> attr = new HashMap<>();
-        attr.put("com.google.appengine.api.users.UserService.user_id_key", "0987");
+        attr.put("com.google.appengine.api.users.UserService.user_id_key", "5678");
```
kirantumkur, line 38: Can you please define a const for this number and also write a comment explain what is that number?

src/main/java/com/google/sps/util/Utils.java
```diff
@@ -1,11 +1,38 @@
 package com.google.sps.util;
 
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
 public class Utils {
+
+    public static final String typeQueue = "queue";
```
kirantumkur, line 8: This should be where the queues are being defined.

src/main/java/com/google/sps/util/Utils.java
henryksloan: `reduce` accumulates collections by repeatedly applying the function in its second parameter to an accumulator variable. The first parameter, `""`, is the initial value of the accumulator, and the function is called by setting actual to each value in `getReader().lines()`. This function just takes the stream of lines from the request body and adds them into one string, so it should probably be called something like `collectRequestLines`.

src/main/java/com/google/sps/servlets/queue/EntityListServlet.java
TheGuyWhoCodes: fixed with 
6182087

src/main/java/com/google/sps/util/Utils.java
TheGuyWhoCodes: what @henryksloan said is correct. i agree that it should be renamed! writing the accumulator this way instead of a traditional for loop saves a lot on complexity, and reads a lot better.

src/main/java/com/google/sps/util/Utils.java
TheGuyWhoCodes: sorry could you explain this more?

src/main/java/com/google/sps/util/Utils.java
TheGuyWhoCodes: Fixed with 
e26a0e2

PR #16, opened 2020-07-08
Add servlet for adding and reading reviews

henryksloan:
Note: The majority of files changed reflect a minor refactor that changed imports

## ReviewServlet
A servlet that provides validated and authenticated database reads and writes for reviews. doGet takes three parameters: contentType {book | movie}, contentId, userId. Valid requests provide either contentType+contentId OR just userId. The request throws 400 if any other combination of parameters is provided, or if the parameters used have invalid, null or empty values.

When querying by userId, 404 is thrown if no user is found with ID == userId. If the request is successful, then all reviews made by the given user will be returned. Pagination will be an easy addition (involving ofy().limit(...).offset(...)), but it is not yet clear what implementation would work best for the interface. An empty list is returned if a valid user has no reviews.

When querying by contentType and contentId, 404 is never thrown. This avoids unnecessary API calls, and won't pose a problem as invalid content requests will be rare and transient. In such cases, an empty list is returned.

## ReviewObject
Reviews are represented with the schema { timestamp, authorId, authorName, **contentType, contentId, reviewTitle, reviewBody, rating, contentTitle, artUrl** }. Only the bolded fields are required to construct a ReviewObject, as the rest are generated by UserService and standard library methods.

## ReviewServletTest
GET: The tests ensure correct error codes for invalid, null and empty parameters. They also ensure successful return given valid parameters
POST: The tests ensure correct error codes for invalid, null and empty parameters, as well as for unauthenticated access. They also ensure that reviews can be posted, and that posted reviews can be retrieved both by userId and by contentId+contentType.

## Refactors and Utilities
- Moved schema objects relevant to user authentication to com.google.sps.model.user. Now, every class in the servlets package is a servlet.
- Created a function static Boolean Utils.mediaItemExists(), made possible by the following:
    - Relocated the API-facing interactions of BookDetailsServlet and MovieDetailsServlet to new, separate functions.
    - Reworked the error throwing and handling models of the details servlets
- A few minor typo fixes in documentation


src/main/java/com/google/sps/model/review/ReviewObject.java
```diff
@@ -0,0 +1,107 @@
...
+                this.artUrl = movie.getPosterPath();
+                break;
+            default:
+                throw new IllegalArgumentException();
+        }
+
+        this.reviewTitle = reviewTitle;
+        this.reviewBody = reviewBody;
+        this.rating = rating;
+    }
```
kirantumkur, line 61: In general practice, constructors show never throw an exception. I would suggest executing all code and validations(eg:book or movie or default) which throw exceptions before calling the constructor and invoking the constructor with parameters which should always succed. (Except the rare case of memory alloc failures)

src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
```diff
@@ -41,22 +59,15 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-        final NetHttpTransport httpTransport;
         try {
-            // Can throw an exception if trusted certificate cannot be established
-            httpTransport = GoogleNetHttpTransport.newTrustedTransport();
+            Volume volume = getDetails(id);
+            response.getWriter().println(gson.toJsonTree(volume));
         }
-        catch (Exception e) {
+        catch (GeneralSecurityException e) {
             response.sendError(HttpServletResponse.SC_BAD_REQUEST);
```
kirantumkur, line 67: Is there something like InternalServerError instead of BAD REQUEST?

src/main/java/com/google/sps/servlets/review/ReviewServlet.java
```diff
@@ -0,0 +1,152 @@
...
+     * Returns error 400 if the query parameters are not in either of these formats
+     * Returns error 400 if a parameter is empty or invalid (e.g. "bok")
+     * Returns error 404 if the given user is not found
+     * Simply returns an empty list if the given media ID does not exist to avoid API call
+     * @param request: expects contentType&contentId OR userId
+     * @param response: returns a JSON list of ReviewObject
+     * @throws IOException
+     */
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
kirantumkur, line 39: Please break apart doGet into smaller methods.

src/main/java/com/google/sps/servlets/review/ReviewServlet.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * doPost() attempts to post a review for a given item from a user
+     * Returns error 400 if any parameters are invalid
+     * Returns error 401 if user is not authenticated
+     * @param request: expects contentType, contentId, reviewTitle, reviewBody, and rating
+     * @param response: returns a JSON string of the review if successful
+     * @throws IOException
+     */
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
kirantumkur, line 92: Please break apart doPost into smaller methods

src/main/java/com/google/sps/util/Utils.java
```diff
@@ -8,4 +13,33 @@ public static Integer parseInt(String s) {
             return null;
         }
     }
+
+    // Leverages servlets to check if an item of type contentType exists with ID == contentId
+    // Returns null if parameters are invalid
+    public static Boolean mediaItemExists(String contentType, String contentId) {
```
kirantumkur, line 19: Why is this in utils? is it being used in multiple places?

src/main/java/com/google/sps/servlets/book/BookDetailsServlet.java
henryksloan: Good point, fixing now

src/main/java/com/google/sps/util/Utils.java
henryksloan: Yes, we intend to use this in EntityListServlet POST requests to match reviews. We decided that POST requests related to non-existent media should be blocked.

src/main/java/com/google/sps/model/review/ReviewObject.java
henryksloan: Good point. I've relocated code with exceptions to new functions in ReviewServlet.

src/main/java/com/google/sps/servlets/review/ReviewServlet.java
henryksloan: Done in ccf8759. I separated authentication, validation and response formulation to several new methods.

src/main/java/com/google/sps/servlets/review/ReviewServlet.java
henryksloan: Done in ccf8759. I separated authentication, validation and response formulation to several new methods.

src/main/java/com/google/sps/util/Utils.java
```diff
@@ -8,4 +13,34 @@ public static Integer parseInt(String s) {
...
+            || contentType == null || contentType.isEmpty()) {
+            return null;
+        }
+
+        try {
+            switch (contentType) {
+                case "book":
+                    new BookDetailsServlet().getDetails(contentId);
+                    return true;
+                case "movie":
```
kirantumkur, line 30: Describe "book" and "movie" as enums and reuse the enums everywhere.

henryksloan:
@kirantumkur It seems like Java's support for String enums isn't optimal for this case, so I used a static class with a nested constructor. That also allowed me to provide a function static boolean isType(). I also moved much of the duplicate code in tests to new functions, and moved some dummy values to constants.

src/main/java/com/google/sps/util/Utils.java
henryksloan: Done in 8dc2575 with a static class.

src/test/java/com/google/sps/servlets/review/ReviewServletTest.java
```diff
@@ -0,0 +1,299 @@
...
+    public static final String DUMMY_REVIEW_TITLE = "Test review";
+    public static final String DUMMY_REVIEW_BODY = "This is a test review";
+    public static final String TOO_BIG_RATING = "7";
+    public static final String GOOD_DUMMY_RATING = "3";
+
+    public static final String DUMMY_USER_ID = "123";
+    public static final String DUMMY_EMAIL = "test@example.com";
+    public static final String DUMMY_USERNAME = "test";
+    public static final String DUMMY_DOMAIN = "example.com";
+    public static final String DUMMY_PROFILE_PIC_URL = "";
```
kirantumkur, line 40: Please camelCase ALL variables.

src/main/java/com/google/sps/util/Utils.java
```diff
@@ -1,11 +1,58 @@
...
 public class Utils {
+    public static class ContentType {
+        public static final String BOOK = "book";
+        public static final String MOVIE = "movie";
+        private static final List<String> types = Arrays.asList(BOOK, MOVIE);
+
+        private ContentType() {}
+
+        public static boolean isType(String s) { return types.contains(s); }
+    }
```
kirantumkur, line 19: Lets have a chat to understand why we cant use enums. This is generally done using enums.

henryksloan: Fixed with b1ca146.

henryksloan:
@kirantumkur I got stuck on that detail for a little bit. It turns out that Enum.valueOf() as shown in that Stack Overflow returns the value associated with that key in the Enum. For example, if the enum was
```
enum Color {
RED(1),
BLUE(3),
GREEN(8);
}
```
Then `Color.valueOf("RED")` would return `1`. `Color.valueOf(1)` would throw `IllegalArgumentException`, as `valueOf` is searching for a matching key, not a matching value. The same principle applies with strings, so there is no standard library way to access keys by value (certainly not in O(1)).

As for making an integral enum like
```
enum ContentType {
BOOK,
MOVIE;
}
```
That would end up mapping BOOK to 0 and MOVIE to 1 (Or some other `Integer`s depending on implementation). Then, when turning a query parameter like `?contentType=book` into a ContentType, I would have to hard-code the string value `"book"`, which defeats the purpose of an enum.

PR #18, opened 2020-07-08
Login button

XLuna24:
Add Login/Register button to navigation bar, which also changes to logout when user is authenticated. Users username appears  in the navigation bar when authenticated.


src/main/webapp/index.html
```diff
@@ -7,9 +7,12 @@
     <script src="home.js"></script>
   </head>
 
-  <body>
+  <body onload="loadLogin()">
       <div class="topnav">
           <h1>Mediaphile</h1>
+          <a id="userName"></a>
+          <a class="button" id="user-login">Login/Register</a>
```
henryksloan, line 14: I still think "Sign in with Google" would be best, especially if you use the official "Sign in with Google" image. At least until we possibly change the login journey so it matches the distinction between register and login.

src/main/webapp/home.css
```diff
@@ -15,6 +15,19 @@ html, body {
     float:left;
 }
 
+#userName {
```
henryksloan, line 18: I think this should be`username` with no caps to match what the api returns.

src/main/webapp/index.html
XLuna24: Updated button to "Sign in with Google" button

src/main/webapp/home.css
XLuna24: fixed

PR #19, opened 2020-07-09
Tests for BookDetailsServlet

henryksloan:
This PR adds tests for BookDetailsServlet

## BookDetailsServletTest
- Ensures no errors are thrown for a GET request to a valid book ID
- Ensures getDetails returns the same book as a direct call to the Books API by checking that the Book IDs match
- Ensures error 400 is thrown for an empty or null ID
- Ensures error 400 is thrown in no book is found with the given ID


TheGuyWhoCodes:
looks good, thanks for deleting the hello world lol

PR #20, opened 2020-07-10
Fix logout callback and occasional BookSearchServlet test failure

henryksloan:
## Logout callback
Before this PR, pressing the logout button would create a popup and occasional side-effects on some machines. This PR fixes this by changing the backend function of the logout button. Instead of using the UserService logout mechanism, the logout button now redirects to the login callback endpoint with the ?logout=1 parameter. That endpoint then removes the cookies associated with the user's login and redirects to the homepage. This removes all unwanted side-effects from logouts, the need to refresh, and the vast majority of the latency associated with logging out.

## Fixed book search tests
Before this PR, the BookSearchServlet test that ensures consistency between the Books API call response and the servlet response would occasionally fail due to API indeterminism. This PR splits `testNormalSearch` into two test: `testNormalSearch` and `testNormalSearchDetails`. `testNormalSearch` simply ensures that a valid search GET request throws no errors. `testNormalSearchDetails` ensures that both the servlet and the API return valid volume lists with matching IDs between their respective first elements. This is achieved by modularizing BookSearchServlet so the action of searching is separate from the doGet request (this modularity improves code clarity as well, at no performance cost).

### DIFF
--- a/.idea/jarRepositories.xml
+++ b/.idea/jarRepositories.xml
@@ -3,11 +3,6 @@
   <component name="RemoteRepositoriesConfiguration">
     <remote-repository>
       <option name="id" value="central" />
-      <option name="name" value="bintray" />
-      <option name="url" value="https://jcenter.bintray.com" />
-    </remote-repository>
-    <remote-repository>
-      <option name="id" value="central" />
       <option name="name" value="Central Repository" />
       <option name="url" value="https://repo.maven.apache.org/maven2" />
     </remote-repository>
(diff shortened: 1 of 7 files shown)


.idea/misc.xml
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="FrameworkDetectionExcludesConfiguration">
+    <file type="web" url="file://$PROJECT_DIR$" />
+  </component>
```
kirantumkur, line 5: what is this for?

.idea/jarRepositories.xml
```diff
@@ -21,5 +16,10 @@
       <option name="name" value="JBoss Community repository" />
       <option name="url" value="https://repository.jboss.org/nexus/content/repositories/public/" />
     </remote-repository>
+    <remote-repository>
+      <option name="id" value="central" />
+      <option name="name" value="bintray" />
+      <option name="url" value="https://jcenter.bintray.com" />
+    </remote-repository>
```
kirantumkur, line 23: what is this library for?

henryksloan: The three files that were updated in the .idea directory (which is just helpful config files for Intellij) were automatic changes made by Intellij. I've reverted the changes, so now all changed files are relevant to the PR.

PR #21, opened 2020-07-10
Content details

XLuna24:
Add Movie and Book details to content page by setting their id to the div element value their in. Later send that id to content page when the div element is clicked on and have the details displayed.


src/main/webapp/home.js
```diff
@@ -51,26 +51,30 @@ function loadBooks() {
...
 
     const title = document.createElement('div');
     title.className = 'title-box';
-    title.innerText = info.title;
+    title.innerText = infoMain.title;
     
     const description = document.createElement('div');
     description.className = 'description-box';
-    description.innerText = info.description ? info.description : info.overview;
+    description.innerText = infoMain.description ? infoMain.description : infoMain.overview;
```
henryksloan, line 71: Javascript actually has cool syntactic sugar for this. You can just do
`description.innerText = infoMain.description || infoMain.overview;`
The value after the `||` is basically a default value. You can refactor `content.value = ...` as well.

XLuna24: That seems simpler, Thanks!

kirantumkur:
@XLuna24 This PR is failing to compile, and hence failing tests. Please review.

henryksloan:
@kirantumkur The build is actually just failing due to the inconsistent test that was patched in my PR #20. If that is merged before this PR, then Xavier can rebase onto that fix and expect a good compile. Other than that test, the build is succeeding.

henryksloan:
@kirantumkur I just discovered that we can re-run checks in the details screen of Github Actions. I just queued the tests to rerun, so they'll most likely be fine.

src/main/webapp/home.js
```diff
@@ -82,6 +86,16 @@ function displayContent(content) {
     document.getElementById(content).style.display = "block";
 }
 
+function contentDetails(id) {
+    let type = "book"
+    if(document.getElementById("movie-list").style.display === "block") {
+        type = "movie";
+    }
+    localStorage.setItem('ident', id);
```
TheGuyWhoCodes, line 94: why use local storage here?

kirantumkur:
@henryksloan @TheGuyWhoCodes Please review this and provide comments or approve this!

src/main/webapp/home.js
XLuna24: That's how I've been able to retrieve the id and type of content when loading content details page. Is there a better way?

XLuna24: done, changed from local storage to query parameters.

PR #25, opened 2020-07-13
Add Struct files and Auth Guard for Future Components

TheGuyWhoCodes:
Part 3

## Majors
 - Adds "struct files" to the project
    - This allows us to map any API requests to actual interfaces, this makes for cleaner code. This is just the base of what we need. I think we will also need support for BookDb objects and MovieDb objects (The Meta Data)
 - Adds Auth Guard
    - For any pages that need auth to reach, this file checks login status from our backend before completing the request. An example of this would be for a user to check the settings page, you should only see this page if the user is authorized in our app.


kirantumkur:
@XLuna24 @henryksloan Please review.

mediaphile/src/app/auth/auth-guard.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    return this.getLoginStatus();
+  }
+
+  async getLoginStatus() {
+    let responseData = this.getStatus().then(authState => {
+      if(authState.loggedIn) {
+        this.loginStatus.setValues(authState.id, authState.url, authState.loggedIn)
+      } else {
+        this.loginStatus.setValues(authState.id, authState.url, authState.loggedIn)
+        this.router.navigate(["/login"]);
```
henryksloan, line 23: So the getLoginStatus function is called by ngOnInit, and will redirect to the login page if the user is not logged in? Shouldn't there be a login redirect button on the front page rather than an automatic redirect?

mediaphile/src/app/struct/MovieSearchResult.ts
```diff
@@ -0,0 +1,6 @@
+export interface MovieSearchResult {
+  results: {}[];
```
henryksloan, line 2: It seems like this construction is equivalent to `results: []`, which makes more sense, since it's just an empty array.

mediaphile/src/app/struct/queue.entity.ts
```diff
@@ -0,0 +1,7 @@
+export interface QueueEntity {
+  id: string
+  title:string
+  type:string
```
henryksloan, line 4: Missing space before type

mediaphile/src/app/struct/MovieSearchResult.ts
TheGuyWhoCodes: This won't work as this is saying we are getting an array of type object. In the future this should be refactored to something like MovieDb[], where MovieDb is another struct with metadata.

mediaphile/src/app/auth/auth-guard.service.ts
TheGuyWhoCodes: so this is done only on a page load while everything is being rendered. so like if a user who isn't authenticated tries to go to a url like `mediaphile.com/home`, the else statement on line 23 would kick in so instead of rendering the page that isn't supposed to be shown to non-authenticated users, it'll instead go to the login page for them to authenticate.

mediaphile/src/app/struct/MovieSearchResult.ts
henryksloan: Oh, I see; that's just a typescript constraint, right?

TheGuyWhoCodes: yup, if you try to access the object within that array without telling angular it in an `{}[]` it will freak out. but this should be switched over to using an actual object, it'll be safer to access.

mediaphile/src/app/struct/queue.entity.ts
TheGuyWhoCodes: fixed with bc686b8

kirantumkur:
@XLuna24 @henryksloan  Please review this and provide comments or approve this!

PR #26, opened 2020-07-13
Add greeting page with login button

TheGuyWhoCodes:
## Majors
 - Adds "greeting" landing page
 - This page is a splash page for people who may not be logged in to still use the site
 - Contains a search bar, and login / register button

## Minors
 - Adds basic creation test. 

## Screenshot
image

### DIFF
--- a/mediaphile/src/app/pages/greeting/greeting.component.spec.ts
+++ b/mediaphile/src/app/pages/greeting/greeting.component.spec.ts
@@ -1,0 +1,32 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { GreetingComponent } from './greeting.component';
+import {Router} from "@angular/router";
+import {HttpClientTestingModule} from "@angular/common/http/testing";
+import {RouterTestingModule} from "@angular/router/testing";
+import {InfoService} from "../../info.service";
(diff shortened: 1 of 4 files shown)


kirantumkur:
@XLuna24 @henryksloan  Please review this and provide comments or approve this!

mediaphile/src/app/pages/greeting/greeting.component.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(GreetingComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
```
henryksloan, line 30: At this point we're just testing for the successful creation of the component, right? What are the eventual capabilities of these tests? Will we be able to verify visual aspects of the page, or just ensure backend requests and processing are working properly?

TheGuyWhoCodes: yup exactly that! so once we finalize layouts etc, then we can assert "is this h2 tag being created" etc. we can also ensure backend endpoints work correctly etc.

PR #30, opened 2020-07-13
Add homepage layout

TheGuyWhoCodes:
## Majors
 - Adds three components, homepage entity, queue entity, and over arching homepage entity 
 - Has tab setup for recent activity from friends
 - This can be refactored at a later time, just a prototype design for now

## Minors
 - Adds tests for the three components

### Without anything in the list
image
### With movie
image
### Mobile
image

### DIFF
--- a/mediaphile/src/pages/home/home-entity/home-entity.component.html
+++ b/mediaphile/src/pages/home/home-entity/home-entity.component.html
@@ -1,0 +1,8 @@
+<div class = "entity-container">
+  <a href ="{{getEntityUrl()}}">
+    <img [src]="getEntityImageUrl()">
+    <span class = "film-title-wrapper">
+        <h3>{{getEntityTitle()}}</h3>
+    </span>
+  </a>
+</div>
--- a/mediaphile/src/pages/home/home-entity/home-entity.component.scss
+++ b/mediaphile/src/pages/home/home-entity/home-entity.component.scss
@@ -1,0 +1,56 @@
+img {
+  width: auto;
+  height: auto;
+  max-width: 75%;
+  max-height: 75%;
+  box-shadow: 2px 2px 2px #bdbdbd;
+  margin-right:1em;
+  transition: 0.3s;
+}
+img:hover {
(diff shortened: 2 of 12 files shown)


henryksloan:
The design looks pretty good for now! We should move to movie/book generic phrasing ASAP. `Queued` is definitely fine, but how about `Read/Watched List` for the middle option for now. That's a bit clunky, but I think it's clearer and less confusing.
The mobile version looks great as well, but I'm not sure about the clipboard for the queued icon. The Read/Watched checkmark is very intuitive, but I think maybe a clock would be better for queue (see "Watch later" on Youtube).

The following can be put off until a UX update:
Also regarding the mobile version: If it's not too hard, maybe we should have bigger buttons for those tabs. Currently very little of the horizontal screen is used, and they seem a bit small for a touchscreen (thumb sized icons are good, according to Apple).
It would also be great to show the text for the currently selected option. It would be good to show that text below the icons, so as not to move the icons while a user is clicking through them.

TheGuyWhoCodes:
> The design looks pretty good for now! We should move to movie/book generic phrasing ASAP. `Queued` is definitely fine, but how about `Read/Watched List` for the middle option for now. That's a bit clunky, but I think it's clearer and less confusing.
> The mobile version looks great as well, but I'm not sure about the clipboard for the queued icon. The Read/Watched checkmark is very intuitive, but I think maybe a clock would be better for queue (see "Watch later" on Youtube).
> 
> The following can be put off until a UX update:
> Also regarding the mobile version: If it's not too hard, maybe we should have bigger buttons for those tabs. Currently very little of the horizontal screen is used, and they seem a bit small for a touchscreen (thumb sized icons are good, according to Apple).
> It would also be great to show the text for the currently selected option. It would be good to show that text below the icons, so as not to move the icons while a user is clicking through them.

Do you think the "Viewed" verb would be better? And for sure I think the icons will need to be changed at some point. I was sticking with FontAwesome but some of their icons are payed for etc. If you find a better one lmk!

PR #31, opened 2020-07-13
Create a "Login Page" for Angular

TheGuyWhoCodes:
## Majors
 - Adds a "Login" page. This login page has a simple button to login, then redirects over to the Google oAuth login system
 - Does this by calling login endpoint to get login oAuth endpoint, then redirects to whatever it gets from the API
 - In the future this should get a check to make sure we aren't already logged in.
## Minors
 - Adds creation test for /login
## Screenshots
image

### DIFF
--- a/mediaphile/src/app/pages/login/login.component.html
+++ b/mediaphile/src/app/pages/login/login.component.html
@@ -4,6 +4,6 @@
   <h2 class="login-header">Log in to Mediaphile</h2>
 
   <div class="login-container">
-    <a (click)="login()"> <img class = "login-button" src="assets/sign-in.png"></a>
+    <a (click)="login()"> <img class="login-button" src="assets/sign-in.png"></a>
   </div>
 </div>
--- a/mediaphile/src/app/pages/login/login.component.html
+++ b/mediaphile/src/app/pages/login/login.component.html
@@ -1,0 +1,9 @@
+<div class="login">
+  <div class="login-triangle"></div>
+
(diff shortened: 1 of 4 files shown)


mediaphile/src/app/pages/login/login.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="login">
+  <div class="login-triangle"></div>
+
+  <h2 class="login-header">Log in to Mediaphile</h2>
+
+  <div class="login-container">
+    <a (click)="login()"> <img class = "login-button" src="assets/sign-in.png"></a>
```
henryksloan, line 7: No space around `=` in attributes

mediaphile/src/app/pages/login/login.component.ts
```diff
@@ -0,0 +1,21 @@
...
+})
+export class LoginComponent implements OnInit {
+
+  constructor(private infoSvc: InfoService) { }
+
+  ngOnInit(): void {
+  }
+
+  public login() {
+    this.infoSvc.login().subscribe(x => {
```
henryksloan, line 17: Rename `x` to something descriptive like loginStatus.

TheGuyWhoCodes: Fixed per 368cfb8

mediaphile/src/app/pages/login/login.component.html
TheGuyWhoCodes: Fixed per 368cfb8

PR #28, opened 2020-07-13
Add conflict safety and media item existence checks to entity list servlet

henryksloan:
## EntityListServlet
- Changed error when the wrong user tries to post a review to 401 (SC_UNAUTHORIZED)
- Added validation for input parameters using mediaItemExists
    - If the item's Id cannot be parsed, then returns error 400 (SC_BAD_REQUEST)
    - If the given media item doesn't exist, returns error 404 (SC_NOT_FOUND)
- If the user already has a review for the given object, returns error 409 (SC_CONFLICT)

## HttpUtils
- Changed sendInvalidPostResponse signature to allow arbitrary error codes

## EntityListServletTest
- Added a test ensuring a conflict error is thrown when a duplicate list item is POSTed

# Major Edit
- **The majority of files changed just reflect several file renames that changed imports**
- Renamed `EntityDbQueue` to `MediaListItem` (i.e. a member of a media list such as Queue or Viewed)
    - Moved `TYPE_` constants to MediaListItem class, as it is the parent of all item types
    - Rename `timeStamp` to `timestamp`, as it is one word
    - The targeted media item is now identified by `mediaType` and `mediaId`
    - The media list item itself is not identified by a field `id`, a UUID which is no longer used in the code
    - The type of list (i.e. viewed/queue) is now identified by a field `listType`
    - `userID` has been renamed to `userId` is compliance with the Google Style Guide definition of camelCase
    - `QueueResponse` has been renamed to `MediaListResponse`
    - `WatchedQueueObject` has been renamed to `QueueListItemObject` (i.e. an Objectify Object representing an Item in the Queue list)
    - `WantToWatchQueueObject` has been renamed to `ViewedListItemObject` (i.e. an Objectify Object representing an Item in the Viewed list)
    - These two objects have also been appropriately renamed in the Datastore by modifying the Objectify @Entity annotation
- Renamed `EntityListServlet` to `MediaListItemServlet` (i.e. a servlet managing items in media lists, such as items in a queue)
    - Added JavaDoc comments to `doGet`, `doPost` and `doDelete`
    - Rename numerous variables to reflect the new `MediaListItem` model
    - **POST requests now accept bodies with the new MediaListItem schema**
    - **DELETE now requires a userId, a mediaType and a mediaId**

PR #33, opened 2020-07-14
Profile frontend

XLuna24:
Add link from search page to user's profile page to view their list of books and movies they wish to see or have seen. Set profile pages to be set to any user. Each profile page contains a box with the user's first letter of their username and their username. Besides this is their queue list which display's movies and books they wish to see. The Watched/Read list is for content the user has already read or watched. Each book and movie entity is displayed in a small box with their title and the type of content(book/movie).


src/main/webapp/profile/profile.js
```diff
@@ -0,0 +1,71 @@
...
+    .then(response => response.json())
+    .then((user) => {
+        document.getElementById("username").innerText = user.username;
+        document.getElementById("profile-pic").innerText = user.username.charAt(0);
+    });
+}
+
+function loadQuery(id) {
+    
+    fetch('/list/entity?userId=' + id + '&entityType=queue')
```
henryksloan, line 19: I have a PR up (#28) that changes this endpoint's parameters. If this wasn't working for you earlier, it may be because `userId` was actually expecting `userID` (capital `D`) before I updated my PR. The `entityType` parameter has also been renamed to `listType`. Similarly, several other fields of the object have changed, so check out my PR to see the new specification for `/list/entity`. You should probably wait for that PR of mine to be merged, then rebase on that to ensure this interface uses the new spec.

src/main/webapp/profile/profile.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function displayContent(content) {
+    document.getElementById("queue-list").style.display = "none";
+    document.getElementById("watched-read").style.display = "none";
+
+    document.getElementById(content).style.display = "block";
+}
+
+function createList(info) {
```
henryksloan, line 46: This function will also likely require changes based on #28. See my other comment.

src/main/webapp/profile/profile.js
henryksloan: @XLuna24 My PR #28 has been merged, so you should fix this up w.r.t. that

PR #35, opened 2020-07-15
Adds Review Support to Angular + Several Bug Fixes

TheGuyWhoCodes:
## Majors
 - Adds a "Review" section to the frontend angular
    - Split up between multiple components. The components are: ReviewEntity , ReviewSubmit, and Review component.
        - The Review Entity is the individual review, gets created with ngFor
        - The Review Submit is the form to submit the review
        - The Review component combines the other two components together
 - Redirects the user to `/home` on successful login INSTEAD of `/`
 - Fixes routing issue in web.xml
- Adds bootstrap visual library to `package.json`

## Minors
 - Removes `app.yaml` routing config, not needed anymore
 - Bug fixes related to #28 

## Screenshots
review1
review2
review3
review4
review5


mediaphile/src/app/app.module.ts
```diff
@@ -21,6 +21,10 @@ import {AuthGuardService} from "./auth/auth-guard.service";
 import {LoginStatus} from "./auth/login.status";
```
henryksloan, line 21: We should go through and make sure we're consistent about spaces within curly brackets (i.e. always include braces like `{ LoginStatus }`). If we fix it in all files, then it should be its own PR. For now, let's just fix it in individual files as they get changed in relevant PRs.

mediaphile/src/app/info.service.ts
```diff
@@ -65,20 +67,41 @@ export class InfoService {
...
+      "userId": userID
     })
   }
 
   public getQueue(userID: string, type: string) {
     return this.http.get(this.postQueueEndpoint, {
       params: {
-        "userID": userID,
-        "entityType": type
+        "userId": userID,
```
henryksloan, line 82: Rename `userID` to `userId`

henryksloan:
@TheGuyWhoCodes I just notice this has a bunch of duplicate PRs. You might have to resolve that with a rebase or something on your local machine, then push with `git push --force` (without force, it'll ask you to pull again and repeat the whole process). You should research the problem to see if there's an easy way, but that's what I had to do both times I encountered this problem.

TheGuyWhoCodes:
> @TheGuyWhoCodes I just notice this has a bunch of duplicate PRs. You might have to resolve that with a rebase or something on your local machine, then push with `git push --force` (without force, it'll ask you to pull again and repeat the whole process). You should research the problem to see if there's an easy way, but that's what I had to do both times I encountered this problem.

The rebase right now looks really bad, lots of commits to sift through, i'll try to find something that can fix this

mediaphile/src/app/pages/helper/review-submit/review-submit.component.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.loading = true;
+    this.infoSvc.postReviewForMedia(this.id, this.type, this.hero.currentRate,
+      this.hero.title, this.hero.review).subscribe(reviewStatus => {
+        this.loading = false
+      this.successfullySubmitted = true;
+    }, error => {
+      this.loading = false
+      this.error = true;
+    })
+  }
```
kirantumkur, line 41: Try reproing some of the error cases in manual testing...

TheGuyWhoCodes: Just added some additional tests for any errors with 21415f2

mediaphile/src/app/app.module.ts
TheGuyWhoCodes: i think this is a stylistic thing. on any auto completes from the IDE, it'll autofill with spaces for any library files like this:

`import { By } from '@angular/platform-browser';`

and for any files user generated:

`import {InfoService} from "../../../info.service";`

PR #37, opened 2020-07-16
Creates recommendations endpoint and tests

TheGuyWhoCodes:
## Majors
 - Creates `/movies/recommendations`
 - Takes in a movieId and returns the first page of recommendations based off the movie
 - throws 400 if the movieId is missing

## Minors
 - Creates tests for endpoint


henryksloan:
This can be closed; superseded by #53

PR #38, opened 2020-07-17
Add a profile page for the Angular frontend

henryksloan:
# Completed so far
## app/pages/user-profile/user-profile.component.ts
- Provides a profile page as shown in the images section
- Contains a header with a generated profile picture and the user's username
    - The profile picture is the first character of the username, capitalized
    - The background color is generated from a hash of the user's ID
    - The character's foreground color is black or white, depending on the brightness of the background color
- Includes fully working queue and viewed list
- Includes placeholder following and followers lists
    - **These should work as soon as:** 
        - The followers backend is merged
        - `info.service.ts` is checked to ensure it matches the backend schema
        - The `follow-entity` component is integrated with the schema 

## app/pages/home/queue/queue.component.ts
- Revised the text shown when a list is empty
    - Will now say "Find something to add" only if the user is viewing their own queue
    - Will say "Nothing added" otherwise
- The queue lists on the home and profile pages now say "Your Queue" iff the user is viewing their own queue, else just "Queue"

# Prepared for completion
## app/pages/user-profile/follow-list/follow-entity/follow-entity.component.*
- Currently a direct copy of `home-entity`
- To be replaced once the data output of the `follow` endpoint is well-defined
- The design and pagination will be discussed leading up to that

# Images
image

Edit: "Watched List" has been changed to "Read/Watched List"

### DIFF
--- a/mediaphile/src/app/info.service.ts
+++ b/mediaphile/src/app/info.service.ts
@@ -21,6 +21,7 @@ export class InfoService {
   private getBookSearch: string = `${this.apiBackendUrl}books/search`;
   private getReviews: string = `${this.apiBackendUrl}reviews`;
   private loginStatus: string = `${this.apiBackendUrl}login/status`;
+  private userEndpoint: string = `${this.apiBackendUrl}user`;
   private postQueueEndpoint: string = `${this.apiBackendUrl}list/entity`;
 
   constructor(private http: HttpClient, private router: Router, private loginStatusService: LoginStatus) {
@@ -64,6 +65,14 @@ export class InfoService {
     this.loginStatusService.sharedUrl.subscribe(x => {
       window.location.href = (x);
     })
+  }
+
+  public getUser(userId: string) {
+    return this.http.get(this.userEndpoint, {
+      params: {
+        "id": userId,
+      }
+    });
   }
 
   public postQueue(posterPath: String, id: String, type: String, title: String, entityType: String, userId: String) {
--- a/mediaphile/src/app/info.service.ts
+++ b/mediaphile/src/app/info.service.ts
@@ -23,6 +23,7 @@ export class InfoService {
   private loginStatus: string = `${this.apiBackendUrl}login/status`;
   private userEndpoint: string = `${this.apiBackendUrl}user`;
   private postQueueEndpoint: string = `${this.apiBackendUrl}list/entity`;
+  private followListEndpoint: string = `${this.apiBackendUrl}follow`;
 
   constructor(private http: HttpClient, private router: Router, private loginStatusService: LoginStatus) {
   }
@@ -112,6 +113,22 @@ export class InfoService {
         "reviewTitle": title,
         "reviewBody": reviewBody,
         "rating": String(rating)
+      }
+    })
+  }
+
+  public postFollow(userId: string, targetId: string) {
+    return this.http.post(this.followListEndpoint, {
+      "userId": userId,
+      "targetId": targetId // TODO: Ensure these names are correct
+    });
+  }
+
+  public getFollowList(userId: string, type: string) {
+    return this.http.get(this.followListEndpoint, {
+      params: {
+        "userId": userId,
+        "type": type // TODO: Ensure these names are correct
       }
     });
   }
--- a/mediaphile/src/app/pages/user-profile/follow-list/follow-entity/follow-entity.component.spec.ts
+++ b/mediaphile/src/app/pages/user-profile/follow-list/follow-entity/follow-entity.component.spec.ts
@@ -1,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { FollowEntityComponent } from './follow-entity.component';
+
+describe('HomeEntityComponent', () => {
+  let component: FollowEntityComponent;
+  let fixture: ComponentFixture<FollowEntityComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ FollowEntityComponent ]
+    })
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(FollowEntityComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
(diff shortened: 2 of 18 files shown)


kirantumkur:
@XLuna24 @TheGuyWhoCodes Please review..

mediaphile/src/app/pages/user-profile/follow-list/follow-list.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  viewerId: string;
+
+  @Input()
+  type: string;
+  public hasResults: boolean = false;
+  public entities: [] = [];
+  constructor(private infoSvc: InfoService) { }
+
+  ngOnInit(): void {
+    this.infoSvc.getFollowList(this.userID, this.type).subscribe(x => {
```
TheGuyWhoCodes, line 27: Instead of using the boolean here, could you do something like this? https://ultimatecourses.com/blog/angular-ngif-async-pipe

mediaphile/src/app/pages/user-profile/follow-list/follow-entity/follow-entity.component.ts
```diff
@@ -0,0 +1,56 @@
...
+  ngOnInit(): void {
+    if (this.entity["type"] != undefined){
+      this.type = this.entity["type"]
+    }
+    console.log(this.entity);
+    console.log(this.getEntityUrl())
+  }
+  public getEntityImageUrl() : string {
+    if(this.entity["artUrl"]) {
+      return "https://image.tmdb.org/t/p/w500" + this.entity['artUrl']
```
TheGuyWhoCodes, line 25: this has been refactored in another PR. This should instead just return `this.entity['artUrl']`

mediaphile/src/app/pages/user-profile/follow-list/follow-entity/follow-entity.component.ts
```diff
@@ -0,0 +1,56 @@
...
+  type: string
+
+  constructor() { }
+
+  ngOnInit(): void {
+    if (this.entity["type"] != undefined){
+      this.type = this.entity["type"]
+    }
+    console.log(this.entity);
+    console.log(this.getEntityUrl())
```
TheGuyWhoCodes, line 21: be sure to get rid of this, thats my fault

mediaphile/src/app/pages/user-profile/follow-list/follow-entity/follow-entity.component.ts
henryksloan: The follow-entity component in this PR is copied straight from the home-entity as a placeholder. The real code is in #39.

mediaphile/src/app/pages/user-profile/follow-list/follow-entity/follow-entity.component.ts
henryksloan: This is copied, temporary code. See my other comment.

mediaphile/src/app/pages/user-profile/follow-list/follow-list.component.ts
henryksloan: As I said in my other comments, this is copied code that is replaced in my newer PR. I still use a similar feature in the new PR, so I'll definitely look into whether I could make this change there.

mediaphile/src/app/info.service.ts
```diff
@@ -104,6 +114,22 @@ export class InfoService {
         "reviewBody": reviewBody,
         "rating": String(rating)
       }
+    })
+  }
+
+  public postFollow(userId: string, targetId: string) {
```
kirantumkur, line 120: Would it be possible to use follower and followee instead?

mediaphile/src/app/info.service.ts
```diff
@@ -104,6 +114,22 @@ export class InfoService {
...
+    return this.http.post(this.followListEndpoint, {
+      "userId": userId,
+      "targetId": targetId // TODO: Ensure these names are correct
+    });
+  }
+
+  public getFollowList(userId: string, type: string) {
```
kirantumkur, line 127: instead of type, can you call it mediaType?

mediaphile/src/app/pages/user-profile/follow-list/follow-list.component.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(FollowListComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
kirantumkur, line 32: I do not understand this. Is this an assert in production code?

mediaphile/src/app/pages/user-profile/follow-list/follow-entity/follow-entity.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(FollowEntityComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
kirantumkur, line 24: Same as other comment, is this an assert?

mediaphile/src/app/info.service.ts
henryksloan: This is meant to match the parameter names of the API, and we're using userId for now because it reflects the names of the UserService methods. That said, we can look at the followers code once we have a PR for that, and if we change it there, it can reflect here. The final names and so on will be those included in the ultimate revision of #39.

mediaphile/src/app/info.service.ts
henryksloan: Same as above, it matters most that this naming matches that of the follower API, but both can be changed once that PR is up.

mediaphile/src/app/pages/user-profile/follow-list/follow-list.component.spec.ts
henryksloan: As we discussed in our meeting: This test uses Angular's testing libraries. According to https://angular.io/guide/testing, these tests are basically just bound to `ng test`, so they'll only be used if we run that or include it in our CI.

mediaphile/src/app/pages/user-profile/follow-list/follow-entity/follow-entity.component.spec.ts
henryksloan: As we discussed in our meeting: This test uses Angular's testing libraries. According to https://angular.io/guide/testing, these tests are basically just bound to `ng test`, so they'll only be used if we run that or include it in our CI.

PR #40, opened 2020-07-20
Book details display

TheGuyWhoCodes:
## Book Display Component
 - Looks very similar to the Movie Details Page with some added enhancements to the way queues are checked

## Modal Component
 - Implements a Modal Component service that renders a popup when you feed in a title and message.
 - Can be extended to the future to change button names etc

## Minors
 - Removes the "sendError" line from doDelete(), will act like a 200 is an error
 - Fixes home queue menu to not cause a fail with poster properties 

image


kirantumkur:
@henryksloan @XLuna24 Please review....

mediaphile/src/app/pages/book-details/book-details.component.ts
```diff
@@ -7,9 +15,134 @@ import { Component, OnInit } from '@angular/core';
 })
 export class BookDetailsComponent implements OnInit {
 
-  constructor() { }
+
+  public miniusCircle = faMinusCircle;
+
+  public userId: string;
+  public id: string;
```
kirantumkur, line 22: what id is this?

PR #41, opened 2020-07-22
Update Movie Details to reflect Book Details

TheGuyWhoCodes:
## Movie Details
 - Updates Movie Details to reflect new visual changes in Book Details
 - This includes:
     - Pop ups
     - Checking if movies are in a list
     - Button supports visual changes if a movie is already in a list

### DIFF
--- a/mediaphile/src/app/pages/movie-details/movie-details.component.ts
+++ b/mediaphile/src/app/pages/movie-details/movie-details.component.ts
@@ -2,6 +2,11 @@ import { Component, OnInit } from '@angular/core';
 import {InfoService} from "../../info.service";
 import {ActivatedRoute} from "@angular/router";
 import {LoginStatus} from "../../auth/login.status";
+import {faMinusCircle} from "@fortawesome/free-solid-svg-icons";
+import {Observable} from "rxjs";
+import {Title} from "@angular/platform-browser";
+import {NgbModal} from "@ng-bootstrap/ng-bootstrap";
+import {ModalComponent} from "../helper/modal/modal.component";
 
 @Component({
   selector: 'app-movie-details',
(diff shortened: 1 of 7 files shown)


mediaphile/src/app/pages/movie-details/movie-details.component.ts
```diff
@@ -10,74 +15,137 @@ import {LoginStatus} from "../../auth/login.status";
...
+  public movieData: {};
+
+  public queue: Object[];
+  public hasQueue: boolean;
+
+  public watched: Object[];
+  public hasWatched: boolean;
+
+  public isLoggedIn: boolean;
+  public isInQueue: boolean;
```
henryksloan, line 34: There should be a get the MediaListItem by a user for a movie, like we talked about with reviews. Once those backend changes are made, this and isInWatched should be replaced with those simpler checks.

mediaphile/src/app/pages/movie-details/movie-details.component.ts
```diff
@@ -10,74 +15,137 @@ import {LoginStatus} from "../../auth/login.status";
...
     }
-    if(this.id != null){
-      this.infoSvc.getMovieDetails(this.id).subscribe(data => {
-        this.entity = data;
-        this.hasResults = true;
+    if(this.movieId != null){
+      this.entity = this.infoSvc.getMovieDetails(this.movieId);
+      this.entity.subscribe(data => {
+        this.movieData = data;
+        this.title.setTitle(`Mediaphile Listing for "${data["title"]}"`)
```
henryksloan, line 67: Profile titles are in the form Mediaphile :: [username]'s profile. Maybe this should be Mediaphile :: [title], or we could change our title scheme. I just think it would be good to have the title early in the string so more of it will show in the browser tab.

PR #44, opened 2020-07-23
Follow servlet

XLuna24:
Add a list of followers and following to Datastore. Returns a list of user objects by sets of 20. Users can follow and unfollow each other


kirantumkur:
@henryksloan @TheGuyWhoCodes Please review

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+    private final UserService userService = UserServiceFactory.getUserService();
+    private final Gson gson = new GsonBuilder().serializeNulls().create();
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    @Override
+    protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        response.setContentType("application/json; charset=utf-8");
+
+        String userId = request.getParameter("userId");
+        String section = request.getParameter("pageNum");
```
henryksloan, line 47: Variable names should be the same or similar to the parameter names. Also, be sure to match this naming with book/movie search: `pageNum` should be `pageNumber`.

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        int page;
+
+        if(userId == null || userId.isEmpty()) {
+                setInvalidGetResponse(response);
```
henryksloan, line 51: Fix indent

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+            return;
+        }
+
+        //page start at 0
+        if(section == null || section.isEmpty()) {
+            page = 0;
+        } else {
+            page = 20 * Integer.parseInt(section);
+        }
```
henryksloan, line 60: `pageNumber` should be the variable from the parameter, and then there should be a new variable here called `startIndex`. Also, the number of results in a page should be a constant `RESULTS_PER_PAGE` (exactly matching `BookSearchServlet.java`).

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+    protected void doDelete(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+        User user = userService.getCurrentUser();
+        if(!userService.isUserLoggedIn()) {
+            sendNotLoggedIn(response);
+            return;
+        }
+
+        String userId = user.getUserId();
+        String followingId = request.getParameter("followingId");
```
henryksloan, line 130: This can be `targetId`. Everywhere else in the code, the following relation is described as `(userId, targetId)`, so I think this could simply match that.

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+import static com.googlecode.objectify.ObjectifyService.ofy;
+
+@WebServlet("/follow")
+public class FollowServlet extends HttpServlet {
+    private final UserService userService = UserServiceFactory.getUserService();
+    private final Gson gson = new GsonBuilder().serializeNulls().create();
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    @Override
+    protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
henryksloan, line 43: We should have `/**` JavaDoc comments on `doGet`, `doPost`, and `doDelete`. Just copy-paste them from `ReviewServlet` or something and change it up, just so we can easily check parameter names and expected errors.

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }
+
+    private List<UserObject> convertToUserObject(JsonElement followList, String type) {
+        JsonArray arrayList = followList.getAsJsonArray();
+        List<UserObject> userObject = new ArrayList<UserObject>();
+
+        for(int i=0; i < arrayList.size(); i++) {
+            String id = (((JsonObject)arrayList.get(0)).get(type)).toString();
+            UserObject user = ofy().load().type(UserObject.class).id(id.substring(1, id.length()-1)).now();
```
henryksloan, line 165: Make some comments here, and split some of the intermediate values into their own variables. I can't quite tell what `((JsonObject)arrayList.get(0)).get(type))` is doing, or why we're using substring.

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+            user.setEmail("");
+            userObject.add(user);
+        }
+        return userObject;
+    }
+
+    public int getListCount(String userId, String type) {
+        int size = 
+        ofy().load().type(FollowItem.class).filter(type, userId).count();
```
henryksloan, line 174: Indent

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        List<FollowItem> following = getList(userId, FollowItem.TYPE_FOLLOWING, page);
+
+        List<UserObject> followerUserObjects = convertToUserObject(gson.toJsonTree(followers), "userId");
+        List<UserObject> targetUserObjects = convertToUserObject(gson.toJsonTree(following), "targetId");
+
+        int numFollowers = getListCount(userId, "targetId");
+        int numFollowing = getListCount(userId, "userId");
+
+        FollowListObject result = new FollowListObject(followerUserObjects, targetUserObjects,
+         numFollowers, numFollowing);
```
henryksloan, line 72: Indent should be 4 spaces.

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+
+    private List<FollowItem> getList(String userId, String followType, int page) {
+        if(followType.equals(FollowItem.TYPE_FOLLOWERS)) {
+            //Sets user as target to retrieve followers.
+            return ofy().load().type(FollowItem.class).limit(20).offset(page).filter("targetId", userId).list();
+        } else {
+            //Sets user as follower to retrieve following.
+            return ofy().load().type(FollowItem.class).limit(20).offset(page).filter("userId", userId).list();
```
henryksloan, line 155: Change this 20 to the same constant I recommended above

src/main/java/com/google/sps/model/follow/FollowListObject.java
```diff
@@ -0,0 +1,44 @@
...
+import com.googlecode.objectify.annotation.Entity;
+import com.googlecode.objectify.annotation.Id;
+import com.googlecode.objectify.annotation.Index;
+import com.google.sps.model.user.UserObject;
+
+import java.util.List;
+
+@Entity
+public class FollowListObject {
+    public FollowListObject() { }
```
TheGuyWhoCodes, line 13: is there a reason to need an empty constructor here?

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
TheGuyWhoCodes: +1

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
TheGuyWhoCodes: +1

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
TheGuyWhoCodes: I agree, 20 as a constant and change the name of section to be something a little more meaningful

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        int numFollowing = getListCount(userId, "userId");
+
+        FollowListObject result = new FollowListObject(followerUserObjects, targetUserObjects,
+         numFollowers, numFollowing);
+
+        response.getWriter().println(gson.toJsonTree(result));
+    }
+
+    @Override
+    protected void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
TheGuyWhoCodes, line 78: same java doc needed

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        response.setContentType("application/json; charset=utf-8");
+        FollowResponse newResponse = new FollowResponse();
+        String body = Utils.collectRequestLines(request);
+
+        User user = userService.getCurrentUser();
+        if(!userService.isUserLoggedIn()) {
```
TheGuyWhoCodes, line 84: should there be a check to see if targetId exists too ie a user should request to follow a userId of -1

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
TheGuyWhoCodes: +1

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+            response.sendError(HttpServletResponse.SC_NOT_FOUND);
+            return;
+        }
+
+        ofy().delete().keys(targetKey).now();
+        response.sendError(HttpServletResponse.SC_OK);
+    }
+
+
+    private List<FollowItem> getList(String userId, String followType, int page) {
```
TheGuyWhoCodes, line 149: javadoc

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        if(followType.equals(FollowItem.TYPE_FOLLOWERS)) {
+            //Sets user as target to retrieve followers.
+            return ofy().load().type(FollowItem.class).limit(20).offset(page).filter("targetId", userId).list();
+        } else {
+            //Sets user as follower to retrieve following.
+            return ofy().load().type(FollowItem.class).limit(20).offset(page).filter("userId", userId).list();
+        }
+    }
+
+    private List<UserObject> convertToUserObject(JsonElement followList, String type) {
```
TheGuyWhoCodes, line 159: javadoc

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+            user.setEmail("");
+            userObject.add(user);
+        }
+        return userObject;
+    }
+
+    public int getListCount(String userId, String type) {
+        int size = 
+        ofy().load().type(FollowItem.class).filter(type, userId).count();
+        return size;
```
TheGuyWhoCodes, line 175: this can just be reduced down to `return ofy().load().type(FollowItem.class).filter(type, userId).count();`

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    private QueryKeys<?> getUserItems(String userId, String targetId) {
```
TheGuyWhoCodes, line 178: javadoc

src/main/java/com/google/sps/model/follow/FollowListObject.java
henryksloan: I think it's required for ObjectMapper, but I'm not 100% sure.

XLuna24: I thought ObjectMapper required it too, but seems to work fine without it.

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
XLuna24: Done

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
XLuna24: Done

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
XLuna24: Yes, thats a good idea. I've added it in with the last push

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
XLuna24: Done

src/main/java/com/google/sps/servlets/follow/FollowServlet.java
XLuna24: Done

henryksloan:
@XLuna24 Just to be clear, we need an optional query parameter for the `GET` request called `targetId`. If that parameter is given, then we'll just return `true` or `false` depending on whether `userId` follows `targetId`. We'll use that to decide whether we should initially show the "follow" or "unfollow" button.

src/main/java/com/google/sps/model/follow/FollowItem.java
```diff
@@ -0,0 +1,54 @@
...
+public class FollowItem {
+
+    public static final String TYPE_FOLLOWERS = "followers";
+    public static final String TYPE_FOLLOWING = "following";
+
+    @JsonProperty
+    @Index
+    @Id
+    private Long id;
+
```
kirantumkur, line 20: Why are you using Long? why not Int? why not string (Eg: UUID)?

@XLuna24 @henryksloan

henryksloan: The `id` field with the `@Id` annotation is an automatically generated ID from Objectify. According to https://github.com/objectify/objectify/wiki/Entities, only `Long` is both nullable and automatically generated.

TheGuyWhoCodes: this ID is generated by datastore and isn't used for anything on our end, just for datastore to index the entity

kirantumkur: But, do we designate it as Long? Or does datastore create the "Long"? What is the max number of items we can store if we use Long? Can we using UUID instead?

henryksloan: Objectify automatically populates the field annotated `@Id`, and it can only be `Long` for our use case. However, the Objectify library abstracts away the generation, so it is guaranteed to be a unique ID. Java `LONG_MAX` is 9,223,372,036,854,775,807.

PR #46, opened 2020-07-24
Fixes Production Routing Issues

TheGuyWhoCodes:
Found a workaround for the `index.html` routing issue.

## Issue
On production builds of the server, the `web.xml` behaves a little differently from what happens locally. I imagine this has to be due with the way Appengine handles servlets. The routing would be fine locally, but would fail on production.

## Fix?
This PR fixes this by rendering out a .jsp version of the HTML file, this renders properly and will require no changes from the Ci / CD or locally building.


TheGuyWhoCodes:
@henryksloan yup, check out the latest prod here: http://kirankubetest2.appspot.com/

PR #39, opened 2020-07-20
Angular followers frontend for user profiles

henryksloan:
**This PR has been updated to reflect #44**
**The only features not included in this PR now are follow list page selection (currently only shows page 0) and follow count display on mobile**

# Done
This PR adds to #38 a follow button and fully working followers/following lists with dummy data
## info.service.ts
- Added dummy users for use in dummy follow lists
- Added utilities to produce observables and requests for `/follow` API
    - POST new follow
    - DELETE existing follow
    - GET full list of type={following/followers}
    - GET boolean whether or not userId follows otherId
    - **These GETs are currently populated with dummy data, but contain draft code in the comments**
## pages/user-profile/follow-list/follow-entity/*
- Replaces placeholder entities from #38
- Displays the user's profile picture and name
- Encompasses a link to the user's profile

# TODO
- Create a default profile picture implementation that can scalably be included in lists like this (the hashing implementation from `UserProfile` should be fine, assuming we have pagination or low user count)
- Add follower/following count in a mobile-friendly manner
    - Possibly use a different display method for these numbers on mobile
- Ensure follow button and lists display well on mobile

# Screenshots
image

PR #47, opened 2020-07-24
Working followers frontend

henryksloan:
This PR contains some final fixes to the followers frontend.
@kirantumkur This only has minor changes from #39, but it needs to be merged for the demo.

## info.service.ts
- Rename some variables to reflect the final names of `/follow` parameters

## Follow list and entities
- Now passes down both userId and userObject
- Add a default profile picture

## user-profile
- Fix deletion and posting of reviews

### DIFF
--- a/mediaphile/src/app/pages/user-profile/user-profile.component.ts
+++ b/mediaphile/src/app/pages/user-profile/user-profile.component.ts
@@ -125,19 +125,10 @@ export class UserProfileComponent implements OnInit {
     let old_followed = this.followed;
     this.followed = !this.followed;
 
-    console.log(old_followed, this.followed, this.profileId, this.userId);
     if (old_followed) {
-      this.infoSvc.deleteFollow(this.profileId).subscribe(x => {
-        console.log(x)
-      }, error => {
-        console.log(error);
-      });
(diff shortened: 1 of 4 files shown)


mediaphile/src/app/pages/user-profile/user-profile.component.ts
```diff
@@ -127,9 +126,9 @@ export class UserProfileComponent implements OnInit {
     this.followed = !this.followed;
 
     if (old_followed) {
-      this.infoSvc.deleteFollow(this.profileId);
+      this.infoSvc.deleteFollow(this.profileId).subscribe();
```
kirantumkur, line 129: what does subscribe do?

henryksloan: The functions `deleteFollow` and `postFollow` in `infoSvc` return `Observable`s, which is basically a Promise of the data returned from the request. Here, `subscribe()` just dispatches the request, and the response goes unused (as these are POST and DELETE requests). Before I added that function call (which I had forgotten), the functions calls `deleteFollow` and `postFollow` just got the observables and didn't dispatch them, so they didn't work.

PR #48, opened 2020-07-24
Implements Recent Activity and Refactors Entity Results

TheGuyWhoCodes:
## Objectify
 - Bumps version up from 4.x -> 5.0. This allows for a needed annotation to have sub classes within POJOs

## Home Entity
 - Refactors to do a tile view instead of list view, looks a lot better imo

## Search Entity
 - Refactors to do a tile view instead of list view, looks a lot better imo

## Recent Activity
 - Implements a "recent activity" section. This endpoint takes in a userId and an offset, and will return a list of "Activity" objects. If there is no activity, it will return an empty list
 - 400 errors are thrown if the query params are missing

## Bugs?
 - Currently, with the way we save entities, there is no way to access username. for right now for any activity that is a "add to queue", it'll show "Someone you follow did XXXXXX", instead of showing username. This will be fixed in a later PR.

### DIFF
--- a/mediaphile/src/app/pages/home/activity/activity-entity/activity-entity.component.ts
+++ b/mediaphile/src/app/pages/home/activity/activity-entity/activity-entity.component.ts
@@ -22,10 +22,8 @@ export class ActivityEntityComponent implements OnInit {
   }
 
   public getArtwork() {
-    if(this.isReview()) {
-      if (this.activity["contentType"] != "book") {
+    if(this.isReview() && (this.activity["contentType"] === "movie")) {
         return "https://image.tmdb.org/t/p/w500" + this.activity["artUrl"]
-      }
     }
     return this.activity["artUrl"];
   }
--- a/mediaphile/src/app/pages/home/activity/activity-entity/activity-entity.component.ts
+++ b/mediaphile/src/app/pages/home/activity/activity-entity/activity-entity.component.ts
@@ -1,0 +1,81 @@
+import {Component, Input, OnInit} from '@angular/core';
+
+@Component({
+  selector: 'app-activity-entity',
+  templateUrl: './activity-entity.component.html',
+  styleUrls: ['./activity-entity.component.scss']
+})
+export class ActivityEntityComponent implements OnInit {
+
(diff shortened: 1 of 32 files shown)


mediaphile/src/app/pages/home/activity/activity-entity/activity-entity.component.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  }
+
+  public isReview(): boolean {
+    return "reviewTitle" in this.activity
+  }
+
+  public getArtwork() {
+    if(this.isReview()) {
+      if (this.activity["contentType"] != "book") {
```
henryksloan, line 26: Shouldn't this just be `== movie`? Also, merge nested if using `||`

src/main/java/com/google/sps/servlets/activity/RecentActivityServlet.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static com.googlecode.objectify.ObjectifyService.ofy;
+
+/**
+ * @author chris
+ * date: 7/21/2020
+ */
```
kirantumkur, line 25: Please remove.

src/main/java/com/google/sps/servlets/review/ReviewServlet.java
```diff
@@ -256,4 +256,39 @@ private boolean validateParameters(String contentType, String contentId,
...
+            default:
+                throw new IllegalArgumentException();
+        }
+        return new String[]{title, artUrl};
+    }
+
+    private ReviewObject createAndSaveReview(UserObject userObject,
+                                             String contentType, String contentId,
+                                             String reviewTitle, String reviewBody, int rating) throws Exception {
+        String[] titleAndArtUrl = getTitleAndArtUrl(contentType, contentId);
```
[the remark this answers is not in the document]
TheGuyWhoCodes: This code was merged from an old master copy, doesn't exist anymore

PR #52, opened 2020-07-28
Adds IsInList Endpoint and Supports Pagination on List Endpoint

TheGuyWhoCodes:
## MediaListItemServlet
 - Adds field "offset" for paginating through queue results, supports up to 24 at a time.
 - Throws 400 if missing

## MediaItemInServlet
 - Servlet that takes a userId and mediaId, returns a two field boolean object that says whether an object is in a queue or watched list
 - Throws 400 if any fields are missing

## Angular frontend
 - Refactors details pages to use new MediaItemInServlet
 - Adds pagination to queue component.

### DIFF
--- a/src/main/java/com/google/sps/servlets/queue/MediaItemInServlet.java
+++ b/src/main/java/com/google/sps/servlets/queue/MediaItemInServlet.java
@@ -1,0 +1,58 @@
+package com.google.sps.servlets.queue;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.sps.model.queue.IsInListResponseObject;
+import com.google.sps.model.queue.QueueListItemObject;
+import com.google.sps.model.queue.ViewedListItemObject;
+
+import javax.servlet.annotation.WebServlet;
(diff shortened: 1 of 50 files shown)


TheGuyWhoCodes:
@henryksloan let me rename that variable before merging. it is in fact a page number, it was originally offset when implementing.

henryksloan:
@TheGuyWhoCodes Alright, that makes sense.

src/main/java/com/google/sps/servlets/queue/MediaItemInServlet.java
```diff
@@ -0,0 +1,63 @@
...
+
+        response.getWriter().println(gson.toJsonTree(listResponse));
+    }
+
+    private boolean isInQueue(String userId, String mediaId) {
+        return !ofy().load().type(QueueListItemObject.class)
+                .filter("userId", userId)
+                .filter("mediaId", mediaId)
+                .order("-timestamp")
+                .list().isEmpty();
```
kirantumkur, line 53: Consider fetching just one item instead of list()

TheGuyWhoCodes: Fixed this with 1a83380. There is a .first() function that works great for this purpose!

PR #53, opened 2020-07-28
Recommendations servlet for books and movies

henryksloan:
## BookSearchServlet
- Now returns total items and total pages in the same format as MovieSearchServlet

## RecommendationsServlet
- Takes required parameters `?mediaType={book | movie}&mediaId={id}`
- Takes an optional parameter `&pageNumber`, zero-indexed, defaulting to zero
- For `mediaType == movie`, uses TMDB API recommendation feature
- **For `mediaType == book`, uses the `associated` feature of the Google Books API**
    - This was hardly documented, but it is luckily exactly what we want
    - Works very similarly to TMDB recommendations
    - **Caveat: ** seemingly only returns 9 results
- Returns an object identical to that returned by `MovieSearchServlet` and `BookSearchServlet`
    - `{ results: [MovieDb or Volume], total_results: int total_pages: int, page: int}`

## RecommendationsServletTest
- Tests appropriate error given empty/null parameters
- Tests non-empty output from valid queries
- Tests non-equal output for pages 0 and 1 of a valid movie query

## Frontend
- Shows total number of results for books as it was previously only shown for movies

### DIFF
--- a/src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
+++ b/src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
@@ -8,6 +8,7 @@ import com.google.api.services.books.model.Volume;
 import com.google.api.services.books.model.Volumes;
 import com.google.gson.Gson;
 import com.google.sps.KeyConfig;
+import com.google.sps.model.results.ResultsObject;
 import org.json.simple.JSONObject;
 
 import javax.servlet.annotation.WebServlet;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/sps/servlets/book/BookSearchServlet.java
```diff
@@ -86,10 +86,8 @@ private JSONObject convertResultsToJson(Volumes searchResults, int pageNumber) {
         JSONObject json = new JSONObject();
```
TheGuyWhoCodes, line 86: Instead of using JSON object, could we just make a Java POJO to represent this return?

src/main/java/com/google/sps/servlets/recommendations/RecommendationsServlet.java
```diff
@@ -0,0 +1,158 @@
...
+                    total, totalPages, pageNumber));
+        }
+        catch (Exception e) {
+            e.printStackTrace();
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+        }
+    }
+
+    private <T> JSONObject convertResultsToJson(List<T> results, int totalResults, int totalPages, int page) {
+        JSONObject json = new JSONObject();
```
TheGuyWhoCodes, line 150: same thing that i said above about converting this to a POJO

henryksloan:
@TheGuyWhoCodes Great idea. I made a POJO that unifies book search results and all recommendation results. I also updated the results page to display the total results returned for books, just like it does with movies.

TheGuyWhoCodes:
LGTM

henryksloan:
@TheGuyWhoCodes I think you have to approve to dismiss the `changes requested`

PR #55, opened 2020-07-28
User search

XLuna24:
Add user search to UserServlet so a user can search other users. Add a normalized username to the username object for better search handling. Add the list to the Angular Search page.

### DIFF
--- a/mediaphile/src/app/info.service.ts
+++ b/mediaphile/src/app/info.service.ts
@@ -186,4 +186,13 @@ export class InfoService {
       }
     })
   }
+
+  public searchUsers(query: string, pageNumber: number) {
+    return this.http.get(this.userEndpoint, {
+      params: {
+        "query": query,
+        "pageNumber": pageNumber.toString()
+      }
+    });
+  }
 }
--- a/mediaphile/src/app/info.service.ts
+++ b/mediaphile/src/app/info.service.ts
@@ -213,6 +213,5 @@ export class InfoService {
       return "https://" + href.substr(7);
     }
     return href;
-
   }
 }
--- a/mediaphile/src/app/info.service.ts
+++ b/mediaphile/src/app/info.service.ts
@@ -50,7 +50,7 @@ export class InfoService {
     return this.http.get(this.getBookSearch, {
       params: {
         "query": query,
-        "pageNumber": page.toString()
+        "pageNumber": (page - 1).toString()
       }
     });
   }
@@ -203,7 +203,7 @@ export class InfoService {
     return this.http.get(this.userEndpoint, {
       params: {
         "query": query,
-        "pageNumber": pageNumber.toString()
+        "pageNumber": (pageNumber - 1).toString()
       }
     });
   }
--- a/mediaphile/src/app/pages/search/results/results.component.html
+++ b/mediaphile/src/app/pages/search/results/results.component.html
@@ -1,10 +1,10 @@
 <div class = "container">
   <h2>Exploring {{query}}: </h2>
   <div>
-    <h6>Found {{total_results}} results!</h6>
+    <h6 *ngIf="!searchResultsContainsNone() && type != 'user'">Found {{total_results}} results!</h6>
     <h6 *ngIf="type=='book'"><img src = "assets/gbs_preview_button1.png"></h6>
-    <div *ngIf="type=='user'" class="row">
-      <div *ngFor="let entry of arrayResults" class="col-md-3">
+    <div *ngIf="type=='user'">
+      <div *ngFor="let entry of arrayResults">
         <app-user-entity [entity]="entry" [userId]="entry['id']"></app-user-entity>
       </div>
     </div>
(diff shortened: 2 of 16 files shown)


mediaphile/src/app/info.service.ts
```diff
@@ -199,10 +199,20 @@ export class InfoService {
...
 
+
+  public searchUsers(query: string, pageNumber: number) {
+    return this.http.get(this.userEndpoint, {
+      params: {
+        "query": query,
+        "pageNumber": pageNumber.toString()
+      }
+    });
+
```
henryksloan, line 210: Missing closing `}`

henryksloan: @XLuna24 I think you need to `git add` `pages/user-entity`

XLuna24: fixed }, and added user-entity

henryksloan: @XLuna24 Thanks for the fix, I'm looking over the check failures now

mediaphile/src/app/pages/search/results/results.component.html
```diff
@@ -1,9 +1,14 @@
 <div class = "container">
   <h2>Exploring {{query}}: </h2>
   <div>
-    <h6>Found {{total_results}} results!</h6>
+    <h6 *ngIf="!searchResultsContainsNone() && type != 'user'">Found {{total_results}} results!</h6>
```
TheGuyWhoCodes, line 4: we shouldn't tack on extra type checks onto an ngIf, instead make a function inside of a component to check this.

mediaphile/src/app/pages/user-entity/user-entity.component.ts
```diff
@@ -18,6 +18,7 @@ export class UserEntityComponent implements OnInit {
   constructor(private infoSvc: InfoService) { }
 
   ngOnInit(): void {
+    console.log(this.entity);
```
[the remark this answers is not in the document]
henryksloan: Good catch, done.

mediaphile/src/app/pages/user-entity/user-entity.component.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  constructor(private infoSvc: InfoService) { }
+
+  ngOnInit(): void {
+    console.log(this.entity);
+    this.hasResults = this.entity != {} && this.entity != undefined;
+    if (this.entity != undefined) {
+      if (this.entity['profilePicUrl'] === "") {
+        this.entity['profilePicUrl'] =
+          "https://3.bp.blogspot.com/-qDc5kIFIhb8/UoJEpGN9DmI/AAAAAAABl1s/BfP6FcBY1R8/s320/BlueHead.jpg";
```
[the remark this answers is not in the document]
henryksloan: Yup, good point

mediaphile/src/app/pages/search/results/results.component.ts
```diff
@@ -77,6 +79,17 @@ export class ResultsComponent implements OnInit {
     });
   }
 
+  private searchUsers(query: string) {
+    this.infoSvc.searchUsers(query, this.pageNumber).subscribe(data => {
+      this.arrayResults.push.apply(this.arrayResults, data);
+      this.hasResults = true;
+
+      if(data === null) {
+        this.canLoadMore = false;
```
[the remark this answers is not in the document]
henryksloan: My bad, I misunderstood your pagination system. I tweaked it so it will work with users and errors as well.

mediaphile/src/app/pages/search/results/results.component.html
henryksloan: You're totally right, fixed.

mediaphile/src/app/info.service.ts
```diff
@@ -199,6 +199,16 @@ export class InfoService {
     })
   }
 
+  public searchUsers(query: string, pageNumber: number) {
+    return this.http.get(this.userEndpoint, {
+      params: {
+        "query": query,
+        "pageNumber": (pageNumber - 1).toString()
```
kirantumkur, line 206: Can the pageNumber be ZERO?

mediaphile/src/app/pages/search/search.component.ts
```diff
@@ -16,6 +16,9 @@ export class SearchComponent implements OnInit {
   public faUsers = faUsers
   public book = "book"
   public movie = "movie"
+  public user = "user"
+
+  public entities: [] = [];
```
kirantumkur, line 21: Can you explain this?

mediaphile/src/app/info.service.ts
henryksloan: The page numbers sent to this function are from the frontend, so they start at 1. If an error or intentional modification causes a negative page number, the backend will just produce an appropriate error.

mediaphile/src/app/pages/search/search.component.ts
henryksloan: These are just aliases for the different content types. Chris used this schema, and I'm just following it, but it basically makes the Angular template code a little cleaner while allowing us to easily change the content types.

TheGuyWhoCodes: Yup exactly what Henry said. Angular needs to know variable types in order to do certain things ie dictionary looksups and iterables, to do this we define type using the : [].

PR #56, opened 2020-07-29
Fix follow button display and add profile pic asset

henryksloan:
- Don't show follow button when user is not logged in
- Add missing profile picture asset


TheGuyWhoCodes:
LGTM

PR #57, opened 2020-07-29
Recommendations angular impl

TheGuyWhoCodes:
## ReccomendationServlet
 - Fixes encoding issue, was encoding as Java name instead of JSON property name

## Entity Component
 - CSS Fixes to make it appropriate for recommendations

## Info Service
 - Add recommendation endpoint in

## Screenshot
image


mediaphile/src/app/info.service.ts
```diff
@@ -199,6 +200,15 @@ export class InfoService {
     })
   }
 
+  public getRecommendations(mediaId: string, mediaType: string) {
+    return this.http.get<{}>(this.getRecommendationsEndpoint, {
```
henryksloan, line 204: This template argument `<{}>` seems redundant. Doesn't it default to `Object` anyway without this argument?

TheGuyWhoCodes: yeah this can be removed all together, doesn't need any type

PR #67, opened 2020-07-30
Recommendations workarounds

henryksloan:
- The system will first attempt to use the original "associated" method for book recommendations
- That method fails quickly when it fails, and the failure will now be caught as an exception, calling the `bookRecommendationsFallback()` function
    - This function takes ownership of the Books API accessor object, speeding up the fallback mechanism
- The fallback will try the following, in order
    - If the book has category labels, match based on the first category by forming a `subject:` query
    - Else, take the first few words of the title (at most 3), and search by those

### DIFF
--- a/src/main/java/com/google/sps/servlets/recommendations/RecommendationsServlet.java
+++ b/src/main/java/com/google/sps/servlets/recommendations/RecommendationsServlet.java
@@ -11,6 +11,7 @@ import com.google.api.services.books.model.Volumes;
 import com.google.gson.Gson;
 import com.google.sps.KeyConfig;
 import com.google.sps.model.results.ResultsObject;
+import com.google.sps.servlets.book.BookDetailsServlet;
 import com.google.sps.util.Utils;
 
 import javax.servlet.annotation.WebServlet;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/sps/servlets/recommendations/RecommendationsServlet.java
```diff
@@ -150,4 +164,51 @@ private void sendBookRecommendations(String bookId, int pageNumber,
...
+            return;
+        }
+
+        // Forms a query
+        // If the book has categories, search by the first category
+        // If the book does not have categories, search by the first few (at most 3) words of the title
+        List<String> categories = volume.getVolumeInfo().getCategories();
+        String query;
+        if (categories != null && categories.size() > 0) {
+            String category = categories.get(0).replaceAll("\\s", "_");
```
TheGuyWhoCodes, line 184: could this be placed into a function?

src/main/java/com/google/sps/servlets/recommendations/RecommendationsServlet.java
```diff
@@ -150,4 +164,51 @@ private void sendBookRecommendations(String bookId, int pageNumber,
...
+            query = "subject:" + category;
+        } else {
+            String[] words = volume.getVolumeInfo().getTitle().split("[\\s-.,!?]+");
```
TheGuyWhoCodes, line 187: same thing here, just in case we need to reuse regex

src/main/java/com/google/sps/servlets/recommendations/RecommendationsServlet.java
henryksloan: Good point, fixed both of these with specialized but general functions. To answer your comment above, yes. The original association feature could eventually work after some IP finagling, and its failure hardly imparts a performance hit. The fallbacks will always be used in production as it stands, but I'm leaving the original to allow for a theoretical fix, as discussed in a meeting.

PR #68, opened 2020-08-03
Greeting component fixup

TheGuyWhoCodes:
# Fixup for the following components:

## Greeting
 - Changes text to be just movies and books instead of books, movies, art, and music
 - Changes search bar width to be a little wider (50%)

## Recent activity
 - Changes recent activity to have no activity text (if someone has nothing in their activity feed)

# Data Creation
## What does this do?
 - The data creation endpoint will generate the following:
   - 100 fake users with fake usernames
   - Generate follow objects for any of these new users, generates about 6-15 follows / followers for the new accounts
   - Generates 250 review for three movies
         - Iron man, Captain Marvel, Ant-Man
   - Generates 250 review for three books
         - 1984, Moby Dick, Steve Jobs


kirantumkur:
@henryksloan @XLuna24 Please review..

